PR #15, opened 2023-05-08
Implement the 'multiple' flag for textfield component

### LINKED ISSUE #7: Implement 'multiple' behaviour
- [x] Default value should change into a list of default values
- [x] Preview component should update UI with button to "add another item"

sergei-maertens:
Closes #7

- [x] Implement form behaviour in preview for `multiple: true`
- [x] Implement `Default value` in configuration form depending on value of `Multiple`

There is extensive context/documentation in the commit messages.

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -1,8 +1,8 @@
+import '@fortawesome/fontawesome-free/css/all.css';
+import '@fortawesome/fontawesome-free/css/v4-shims.css';
 import 'bootstrap/dist/css/bootstrap.min.css';
 import 'formiojs/dist/formio.builder.css';
 import 'react-tooltip/dist/react-tooltip.css';
-
-import '@fortawesome/fontawesome-free/css/all.css';
 
 import {BuilderContextDecorator, ModalDecorator} from './decorators';
 import {reactIntl} from './reactIntl';
--- a/src/components/ComponentPreview.tsx
+++ b/src/components/ComponentPreview.tsx
@@ -1,9 +1,15 @@
 import clsx from 'clsx';
+import {Formik} from 'formik';
 import React, {useState} from 'react';
 import {FormattedMessage} from 'react-intl';
 
 import REGISTRY from '@/registry';
 import {ExtendedEditFormComponentSchema, JSONType} from '@/types';
+
+/*
+  Generic JSON Preview
+  -> move to utils?
+ */
 
 interface JSONPreviewProps {
   data: JSONType;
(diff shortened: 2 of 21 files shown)


.storybook/preview.js
```diff
@@ -1,9 +1,9 @@
+import '@fortawesome/fontawesome-free/css/all.css';
+import '@fortawesome/fontawesome-free/css/v4-shims.css';
```
sergei-maertens, line 2: we're using those in the backend project and I'd like to keep the icons the same for now

src/components/ComponentPreview.tsx
```diff
@@ -16,18 +29,116 @@ const JSONPreview: React.FC<JSONPreviewProps> = ({data, className = ''}) => (
...
+  component: ExtendedWellBehavedComponentSchema;
+  as: React.FC<ComponentPreviewProps>;
+  defaultValue: unknown;
 }
 
+const isWellBehavedComponent = (
+  component: ExtendedEditFormComponentSchema
+): component is ExtendedWellBehavedComponentSchema => !!component.type && !!component.key;
+
+const PreviewMultiple: React.FC<PreviewMultipleProps> = ({component, as: Nested, defaultValue}) => {
```
sergei-maertens, line 46: TODO: generalize this so it also works in the configuration form -> put it in the utils or containers

src/components/ComponentPreview.tsx
```diff
@@ -40,18 +151,51 @@ const ComponentPreview: React.FC<ComponentPreviewProps> = ({component}) => {
...
-            <JSONPreview data={component} />
-          )}
-        </div>
+        <Formik
+          enableReinitialize
+          initialValues={initialValues}
+          onSubmit={() => {
+            throw new Error("Can't submit preview form");
+          }}
+        >
```
sergei-maertens, line 160: this makes a lot more sense to handle in a generic way, rather than burdening every component type with wrapping it themselves. It's also needed for proper multiple false/true toggle behaviour.

svenvandescheur: Shouldn't this show some visual feedback for the user instead of just raising an error?

src/components/formio/multiple.tsx
```diff
@@ -0,0 +1,137 @@
...
+                </tr>
+              ))}
+              <tr>
+                <td colSpan={2}>
+                  <button
+                    type="button"
+                    className="btn btn-primary formio-button-add-another"
+                    onClick={() => arrayHelpers.push(defaultValue)}
+                  >
+                    <i className="fa fa-plus"></i>
```
svenvandescheur, line 95: Do we use font awesome's auto accessibility here or do we need to add aria-hidden?

src/components/formio/component.tsx
```diff
@@ -30,16 +30,11 @@ const Component: React.FC<ComponentProps> = ({
...
-          {tooltip && ' '}
-          <Tooltip text={tooltip} />
-        </label>
-      ) : null}
+      {label && (
+        <ComponentLabel label={label} required={required} htmlId={props.htmlId} tooltip={tooltip} />
+      )}
       {children}
       <div className="formio-errors invalid-feedback">
         {errors.map((error, index) => (
```
svenvandescheur, line 40: Unrelated to this PR but index doesn't seem to be the best key?

svenvandescheur (commented):
Some small suggestions, and can you check if it's possible to replace the index usage with something else? Also: as noted. The readme can be improved by adding a section on compile messages.

src/components/formio/multiple.tsx
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  return (
+    <RenderContext.Provider value={{bareInput: true}}>
+      {label && <ComponentLabel label={label} required={required} tooltip={tooltip} />}
+      <FieldArray name={name}>
+        {arrayHelpers => (
+          <table className="table table-bordered">
+            <tbody>
+              {(fieldValue || []).map((_, index) => (
```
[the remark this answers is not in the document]
sergei-maertens: Yes - this code is basically adapted from the Formik docs itself: https://formik.org/docs/api/fieldarray

src/components/formio/multiple.tsx
sergei-maertens: we don't, as far as I know. This is just replicating the formio vanilla builder, but it can't hurt to add the aria attribute, so I'll do that. Though I think you are particularly masochistic if you try to use the form builder interface through a screen reader :stuck_out_tongue:

src/components/ComponentPreview.tsx
```diff
@@ -16,18 +22,38 @@ const JSONPreview: React.FC<JSONPreviewProps> = ({data, className = ''}) => (
...
+
+  TODO: I think the types currently require this to be defined, so should this even
+  exist? Maybe we should just throw an exception if you're trying to render a component
+  type that's not in the registry?
+ */
 
 const Fallback: React.FC<ComponentPreviewProps> = ({component}) => <JSONPreview data={component} />;
 
-type PreviewState = 'rich' | 'JSON';
+/*
```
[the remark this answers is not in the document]
sergei-maertens: yeah I'll take a look! the primary goal was to add structure to the code itself since there are some logical groups identifiable, but the docs tab itself can benefit from this.

src/components/formio/multiple.tsx
```diff
@@ -0,0 +1,137 @@
+/**
```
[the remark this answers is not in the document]
sergei-maertens: This component is an implementation detail and as such not documented directly in Storybook, so moving the docblock around to the component doesn't result in anything, storybook-wise. I'm leaving it here as it documents the intended usage of the module without burying it deeper in the code.

The behaviour itself is documented and tested in more concrete component stories/variants, as that is the intended public API.

svenvandescheur (approved):

PR #26, opened 2023-07-10
[OF#3215] Add configuration for prefill identifier role

SilviaAmAm:
Fixes open-formulieren/open-forms#3215

### DIFF
--- a/src/components/builder/prefill/identifierRole.tsx
+++ b/src/components/builder/prefill/identifierRole.tsx
@@ -1,10 +1,22 @@
-import {FormattedMessage, useIntl} from 'react-intl';
+import {FormattedMessage, defineMessage, useIntl} from 'react-intl';
 
 import Select from '@/components/formio/select';
 
 const IDENTIFIER_ROLE_OPTIONS = [
-  {label: 'Main', value: 'main'},
-  {label: 'Authorised person', value: 'authorised_person'},
+  {
+    label: defineMessage({
+      description: 'Label identifier role main',
+      defaultMessage: 'Main',
+    }),
+    value: 'main',
+  },
+  {
+    label: defineMessage({
+      description: 'Label identifier role authorised person',
+      defaultMessage: 'Authorised person',
+    }),
+    value: 'authorised_person',
+  },
 ];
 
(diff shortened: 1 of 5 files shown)


src/components/builder/prefill/identifierRole.tsx
```diff
@@ -0,0 +1,34 @@
+import {FormattedMessage, useIntl} from 'react-intl';
+
+import Select from '@/components/formio/select';
+
+const IDENTIFIER_ROLE_OPTIONS = [
+  {label: 'Main', value: 'main'},
+  {label: 'Authorised person', value: 'authorised_person'},
```
sergei-maertens, line 7: the labels should be translatable with react-intl, no?

src/components/builder/prefill/prefill.stories.tsx
```diff
@@ -130,3 +130,37 @@ ChangingPluginClearsAttribute.play = async ({canvasElement}) => {
...
+        attribute: 'plugin-2-attribute-1',
+        identifierRole: 'main',
+      },
+    },
+  },
+};
+ChangingIdentifierRole.play = async ({canvasElement}) => {
+  const canvas = within(canvasElement);
+
+  await waitFor(async () => {
```
sergei-maertens, line 149: I don't think we need the `waitFor` here because the options are not retrieved async but sync, while the prefill plugins and attributes are retrieved from an API endpoint.

src/types/schemas.d.ts
```diff
@@ -12,6 +12,7 @@ export interface DisplayConfig {
 export interface PrefillConfig {
   plugin: string | null;
   attribute: string | null;
+  identifierRole: string;
```
sergei-maertens, line 15: I think this can/should be a union of literals? `'main' | 'authorised_person'`

sergei-maertens (changes requested):

src/components/builder/prefill/prefill.stories.tsx
SilviaAmAm: Mhmm if I remove it, the interaction tests fail :thinking:

sergei-maertens (approved):

PR #34, opened 2023-09-01
Default value + multiple toggle handling in preview

### LINKED ISSUE #21: Handle changes in defaultValue in builder form
These should ensure the component preview gets updated appropriately

Essentially, the Formik `initialValues` should be sourced from this value

Tangentially related - the defaultValues now behaves incorrectly in the builder form when toggling multiple true/false.

- [x] Fix preview
- [x] Fix `defaultValue` i.c.w. `multiple`

sergei-maertens:
Closes #21 

This now has the same interaction as form.io's builder/preview itself.

demo

### DIFF
--- a/src/registry/textfield/textfield-component.stories.ts
+++ b/src/registry/textfield/textfield-component.stories.ts
@@ -36,6 +36,7 @@ export const PreviewInitialValues: Story = {
   name: 'defaultValue in preview',
   play: async ({canvasElement, step}) => {
     const canvas = within(canvasElement);
+    const editForm = within(canvas.getByTestId('componentEditForm'));
     const preview = within(canvas.getByTestId('componentPreview'));
 
     await step('Render single defaultValue in preview', async () => {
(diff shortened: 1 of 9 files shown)


src/registry/textfield/textfield-component.stories.ts
```diff
@@ -0,0 +1,73 @@
...
+      const defaultValue = canvas.getByLabelText('Default Value');
+      await userEvent.type(defaultValue, 'Show me in preview!');
+
+      const previewInput = preview.getByRole('textbox');
+      expect(previewInput).toHaveDisplayValue('Show me in preview!');
+    });
+
+    await step('Toggle to multi=true', async () => {
+      await userEvent.click(canvas.getByLabelText('Multiple values'));
+      // the single defaultValue should become the first element in the multiple array
```
SilviaAmAm, line 52: As discussed in the office, this doesn't seem to work on firefox, but will be tackled in a separate issue

SilviaAmAm (approved):

PR #39, opened 2023-10-10
Store translations on the component

sergei-maertens:
:boom: breaking change!

Partly fixes open-formulieren/open-forms#2958

The new component translations manager now stores the translations directly on each component, in the format as expected by the backend (see open-formulieren/open-forms#3527).

This is not backwards compatible, and trying to open an existing component with legacy format translations will either crash or not display the translations (I suspect the latter).

The types enforce that you specify all component-translatable properties, which are taken from the `@openformulieren/types` package.

Future PRs will add:

- more builder form implementations for different component types
- handle select/radio/selectboxes option label translations


src/components/builder/i18n.stories.tsx
```diff
@@ -21,83 +21,91 @@ export default {
...
-      expect(literal2Inputs).toHaveLength(1);
-    });
-
-    // Enter a value for the nested translatable field
-    await userEvent.type(canvas.getByLabelText('Nested translatable field'), 'Literal 3');
-    await waitFor(async () => {
-      const literal3Inputs = canvas.queryAllByDisplayValue('Literal 3');
-      expect(literal3Inputs).toHaveLength(2);
-    });
+    expect(translationField2).toBeNull();
```
CharString, line 109: This re-evaluates the query?

CharString: What was wrong with the previous interactions? I thought we only change where translations are stored. So I kinda only expect changes in the setup of the canvas.

src/components/builder/i18n.tsx
```diff
@@ -1,173 +1,159 @@
...
 
-import {DataGrid, DataGridRow, Tab, TabList, TabPanel, Tabs, TextField} from '..//formio';
+import {Component, TextField} from '../formio';
+import ComponentLabel from '../formio/component-label';
 
-export type SchemaKey<S> = keyof S & string;
+type ExtractTranslatableProperties<T> = T extends OFExtensions<infer TK> ? TK : never;
+type StringValueProperties<S> = S extends AnyComponentSchema
+  ? keyof FilterByValueType<S, string | undefined>
+  : never;
```
CharString, line 18:
```suggestion
type TranslatableProperties<T> = T extends OFExtensions<infer TK> ? TK : never;
type StringValueProperties<S> = S extends AnyComponentSchema
  ? keyof FilterByValueType<S, string | undefined>
  : never;
```

If I understand correctly, of some FormIO dict T/S these represent the keys that are translatable/Optional[str] and they can be inferred to the bottom type `never`, to make sure dicts that have nu such keys, don't typecheck.

The StringValueProperties I can follow.

But the TranslatableProperties look like they depend on the same type variable name `TK` used in `{openForms?: { translations: ComponentTranslations<TK>}, ...}`. But they're different TK's. And since all keys in `OFExtensions` are marked optional, I don't understand how this type works.

package.json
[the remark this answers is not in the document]
sergei-maertens: I checked and I think they are indeed unused. When replacing them for deep-merge I wasn't sure if there would not be any need at all anymore, but a couple PRs later I haven't needed them.

I'll remove them, they don't affect bundle size/compile times at all but it's stupid and confusing to install unused dependencies.

src/registry/number/edit.tsx
```diff
@@ -116,7 +110,17 @@ const EditForm: EditFormDefinition<NumberComponentSchema> = () => {
...
       <TabPanel>
-        <Translations.ComponentTranslations />
+        <Translations.ComponentTranslations<NumberComponentSchema>
+          propertyLabels={{
+            label: intl.formatMessage(LABELS.label),
+            description: intl.formatMessage(LABELS.description),
+            tooltip: intl.formatMessage(LABELS.tooltip),
+            suffix: intl.formatMessage({
+              description: "Component translations 'suffix' property label",
+              defaultMessage: 'Suffix (kind of value indicator)',
```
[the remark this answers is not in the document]
sergei-maertens: I'll go with the latter, I find it hard to phrase this in simple language.

src/components/builder/i18n.stories.tsx
sergei-maertens: > This re-evaluates the query?

yeah you can re-use references, in playwright too it uses a locator expression which when combined with the `expect` only then "evaluates" the locator

sergei-maertens: > What was wrong with the previous interactions? I thought we only change where translations are stored. So I kinda only expect changes in the setup of the canvas.

The previous setup was a table mapping from literal -> translation without any accessible labels. Now the table has three columns with proper labels, a reference for the value currently defined in the form and a field for the translation.

src/components/builder/i18n.tsx
sergei-maertens: Yeah so `S` is essentially "Formio **S**chema". `T` has a similar target, but it's not bound to be assignable to `AnyComponentSchema` because not all component types support translations (columns for example is purely layout, so a generic type here would break the entire component and I prefer building everything against `AnyComponentSchema`).

I prefer the name `ExtractTranslatableProperties` because it does exactly that: figure out which properties are translatable from the given (concrete) schema. E.g.:

```ts
type F = ExtractTranslatableProperties<TextFieldComponentSchema>; // 'label' | 'description' | 'tooltip' | 'defaultValue' | 'placeholder';
type G = ExtractTranslatableProperties<ContentComponentSchema>; // 'html' ;
```

When passing in the union `AnyComponentSchema` you will indeed get a union of all possible translatable properties, but that's not how these components/types are supposed to be used - when you use the component, you need to pass in the concrete schema to really benefit from the type checking.

In the type definitions for each schema, we explicitly pass in which keys can be translated, see for example: https://github.com/open-formulieren/types/blob/b7e6931fbe7fd0d8abe773dde1f69ea82cf9296b/src/formio/components/textfield.ts#L4

> and they can be inferred to the bottom type never, to make sure dicts that have nu such keys, don't typecheck.

Yes, I think? An example of this would be the `columns` formio component type - nothing is user-facing there, so nothing is translatable.

> But the TranslatableProperties look like they depend on the same type variable name TK used in `{openForms?: { translations: ComponentTranslations<TK>}, ...}`.

The optional keys inside aren't relevant - remember that the `extends` meaning is "can be assigned to" and the optional/`undefined` keys make that assignability a little bit easier.

What's then happening, for example with `TextFieldComponentSchema`, is that you see that that component is created via the [`InputComponentSchema`](https://github.com/open-formulieren/types/blob/b7e6931fbe7fd0d8abe773dde1f69ea82cf9296b/src/formio/components/textfield.ts#L6) primitive, and that itself [extends](https://github.com/open-formulieren/types/blob/b7e6931fbe7fd0d8abe773dde1f69ea82cf9296b/src/formio/base.ts#L148) `OFExtensions<TK>`. So, typescript is able to `infer TK` because it sees `OFExtensions<'label' | 'description' | 'tooltip' | 'defaultValue' | 'placeholder'>`, meaning it infers `TK` to be `'label' | 'description' | 'tooltip' | 'defaultValue' | 'placeholder'`

If you lift this up one level to make it more generic, it distributes this logic over all union members: `TextFieldComponentSchema`, `NumberComponentSchema`... to get the full union for `AnyComponentSchema`. I found this to be the only way to achieve this (with the conditional type) because `keyof AnyComponentSchema['openForms']['translations'][SupportedLocales]` seems to express the same, excep that this takes the intersection of all union members which ends up with `never` (because ultimately nothing intersects with the `html` property from the content component).

CharString (approved):

PR #43, opened 2023-10-20
Add minimal edit-json mode to builder

### LINKED ISSUE #42: Allow editing JSON
For @LaurensBurger

sergei-maertens:
Closes #42

I don't know how to approach the autogenerated component key from the label :( It's tracked on the form field itself whether it was set manually or automatically, so the formik `setValues` doesn't know or has no way communicate that the key from the JSON should be taken.

### DIFF
--- a/src/components/ComponentEditForm.stories.tsx
+++ b/src/components/ComponentEditForm.stories.tsx
@@ -1,4 +1,6 @@
+import {expect} from '@storybook/jest';
 import {Meta, StoryFn, StoryObj} from '@storybook/react';
+import {userEvent, within} from '@storybook/testing-library';
 
 import ComponentEditForm from './ComponentEditForm';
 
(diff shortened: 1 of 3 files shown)


CharString (commented):
I did `npm ci` and `npm storybook` and the Interactions on `Component edit form > Edit JSON` are failing for me in both Firefox and Chromium. But differently.

Firefox:
Firefox

Chromium:
Chromium

sergei-maertens:
> I did `npm ci` and `npm storybook` and the Interactions on `Component edit form > Edit JSON` are failing for me in both Firefox and Chromium. But differently.

They were quite flaky with Brave for me too - I found that if the browser window doesn't have focus, then the assertion tends to fail.

I'll take a look if I can make this more reliable, but I'm starting to suspect that it's react or the browser throttling updates somewhere :thinking:

PR #44, opened 2023-10-24
:construction_worker: Run storybook tests in the three major browsers

sergei-maertens:
Firefox is flaky locally, let's see if that's also the case in CI and fix it.


CharString (approved):
Is the intent to eventually change this back to user-event? Or are we fine with firing just the click event?

PR #47, opened 2023-10-31
✨ Add `checkbox` component

Viicos:
I can't test the storybook locally as running build/watch crashes my computer

### DIFF
--- a/src/components/formio/checkbox.stories.tsx
+++ b/src/components/formio/checkbox.stories.tsx
@@ -37,6 +37,13 @@ export const WithToolTip: Story = {
   },
 };
 
+export const WithDescription: Story = {
+  args: {
+    label: 'With description',
+    description: 'A description',
+  },
+};
+
 export const WithErrors: Story = {
   args: {
     label: 'With errors',
--- a/src/components/formio/checkbox.tsx
+++ b/src/components/formio/checkbox.tsx
@@ -4,21 +4,31 @@ import {Field, FormikHandlers, useFormikContext} from 'formik';
 import {useValidationErrors} from '@/utils/errors';
 
 import Component from './component';
+import Description from './description';
 import Tooltip from './tooltip';
 
 export interface CheckboxProps {
   name: string;
   label?: React.ReactNode;
+  required?: boolean;
   tooltip?: string;
+  description?: string;
   onChange?: FormikHandlers['handleChange'];
 }
 
-const Checkbox: React.FC<CheckboxProps> = ({name, label, tooltip = '', onChange}) => {
+const Checkbox: React.FC<CheckboxProps> = ({
+  name,
+  label,
+  required = false,
+  tooltip = '',
+  description = '',
+  onChange,
+}) => {
   const {getFieldProps} = useFormikContext();
   const {onChange: formikOnChange} = getFieldProps(name);
   const {hasErrors} = useValidationErrors(name);
   return (
-    <Component field={name} type="checkbox">
+    <Component field={name} required={required} type="checkbox">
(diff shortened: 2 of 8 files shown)


src/components/formio/checkbox.stories.tsx
```diff
@@ -37,6 +37,13 @@ export const WithToolTip: Story = {
   },
 };
 
+export const WithDescription: Story = {
+  args: {
+    label: 'With descipriotn',
```
sergei-maertens, line 42:
```suggestion
    label: 'With description',
```

src/components/formio/checkbox.tsx
```diff
@@ -4,21 +4,31 @@ import {Field, FormikHandlers, useFormikContext} from 'formik';
...
 import Component from './component';
+import Description from './description';
 import Tooltip from './tooltip';
 
 export interface CheckboxProps {
   name: string;
   label?: React.ReactNode;
+  required?: boolean;
   tooltip?: string;
+  description?: string;
```
sergei-maertens, line 15: may also be a `React.ReactNode`, right?

src/registry/checkbox/edit-validation.ts
```diff
@@ -0,0 +1,15 @@
+import {IntlShape} from 'react-intl';
+import {z} from 'zod';
+
+import {buildCommonSchema} from '@/registry/validation';
+
+// undefined (optional) for unspecified, otherwise a finite numeric value. Note that
```
sergei-maertens, line 6: please update the comments when you copy paste :-)

src/registry/checkbox/edit-validation.ts
```diff
@@ -0,0 +1,15 @@
...
+// null would be nicer, but formio's schema does not support null for validate.min,
+// validate.max or defaultValue
+const booleanSchema = z.boolean().optional();
+
+const defaultValueSchema = z.object({defaultValue: booleanSchema});
```
sergei-maertens, line 11: If you set `EditForm.defaultValues = {...}` appropriately, I think it'll be pinned to either `false` or `true` and will add in any missing `defaultValue` key.

I think we can just stick to `buildCommonSchema` for checkbox component  - the UI control is a checkbox anyway which forces boolean value.

sergei-maertens (changes requested):
There's an ESM build error to it seems, check if you can reproduce with `npm run build`

src/components/formio/checkbox.tsx
Viicos: It is used this way: `{description && <Description text={description} />}` (and in other components as well)

src/registry/checkbox/edit-validation.ts
Viicos: So it resutls in `const schema = (intl: IntlShape) => buildCommonSchema(intl);`?

src/components/formio/checkbox.tsx
sergei-maertens: yeah, with:

```tsx
interface DescriptionProps {
  text: React.ReactNode;
}
```
but I see it's indeed inconsistent in existing components, so let's leave it a narrow type and widen it if needed.

src/registry/checkbox/edit-validation.ts
sergei-maertens: yep!

At some point we should set up the `useZodErrorMap` hook from the SDK here too, but it's still a bit mixed to set up more specific error messages that need to be translatable. And we need to check how many translations overlap, potentially that should also become its own library.

sergei-maertens (approved):

PR #48, opened 2023-11-02
Implement builder component for options

### LINKED ISSUE #12: Implement observing nested translatable fields
This is relevant for the choices with the `select`, `radio`, `checkboxes` components, amongst others

sergei-maertens:
Closes #12 

This replicates the vanilla formio table. Translation management of options will be tackled in another PR.


Viicos:
image

It seems you can still click on the bottom arrow while pointing on the upper one. Same for the last row:

image

sergei-maertens:
@Viicos I've replaced the sort icon with chevrons - the sort icons have their icon part of the glyph aligned to the top/bottom which creates transparent space between them. However, this seriously messes with their offset/spacing and size and apparently causes the mouse-hover inconsistencies. The chevrons don't have this issue, but they're a little less nice in appearance.

Viicos:
> but they're a little less nice in appearance.

Actually I like it better this way, feels a bit more "modern"

sergei-maertens:
> > but they're a little less nice in appearance.
> 
> Actually I like it better this way, feels a bit more "modern"

yes and no - I'm a bit concerned that they overlap too much with expand/collapse icons. But it'll do for now, let's see what Demi comes up with in the new design.

src/types/paths-test.tsx
sergei-maertens, line 1: TODO: delete this file

Viicos:
Is the _Value derived from label_ story working as expected? I end up with:
image

src/types/paths.ts
```diff
@@ -0,0 +1,88 @@
...
+ */
+export type Paths<T> = T extends T
+  ? {
+      [K in keyof T]-?: K extends string
+        ? ExcludeBareObject<T[K]> extends object
+          ? K | `${K}.${Paths<ExcludeBareObject<T[K]>>}`
+          : K
+        : never;
+    }[keyof T]
+  : never;
```
Viicos, line 70: Seems to work if `T` is defined as `{[K: string]: ...}` (then `Paths<T>` is `string`)

However I've tried the following with `PathsForValueType`:

```ts
type Test = PathsForValueType<{[K: string]: Option[]}, Option[]>
// Test is string | number?
```

But haven't investigated further, still need to understand the other types below.

src/components/builder/values/ValuesTable.stories.ts
```diff
@@ -0,0 +1,159 @@
+import {expect} from '@storybook/jest';
+import {Meta, StoryObj} from '@storybook/react';
+import {userEvent, within} from '@storybook/testing-library';
+
+import {withFormik} from '@/../.storybook/decorators';
+
+import ValuesTable from './ValuesTable';
+
+const ValuesTableComponent = ValuesTable<{
+  values: Option[];
```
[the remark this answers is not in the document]
sergei-maertens: Huj, weird that it builds

sergei-maertens:
Sorry for the resolved conversations, I really should stop typing comments on mobile since GH is pretty bad then. I keep accidentally pressing the buttons and it then discards my draft comment :(

src/components/builder/values/ValuesTable.tsx
```diff
@@ -0,0 +1,109 @@
...
+                    />
+                  </span>
+                </th>
+              </tr>
+            </thead>
+
+            <tbody>
+              {options.map((_, index) => (
+                <tr key={index}>
+                  <OptionRow name={name} index={index} arrayHelpers={arrayHelpers} />
```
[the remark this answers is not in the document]
sergei-maertens: so we're interested in both `name` and `index` for a single option:

1. `name` combined with `index` determines the values namespace for each option, leading to `${name}[${index}].label` and `${name}[${index}].value`
2. the `index` is used in combination with `FormikValues[$name]`, which contains the entire list of all the options for the component - to check whether an option is the last or first element, as that possibly disables one of the sort controls
3. `name` is the only part that is dynamic about this, in practice it's `"data.values" | "values"`, the rest is fixed because of the `Option[]` type.

alternatively we'd need to calculate `isFirst` and `isLast` in the parent component and pass that down to the `OptionRow`, and then the option row prop types get a bit weird since if you have > 2 options, isLast/isFirst cannot both be true at the same time, I picked this approach because it allows you to pass the atomic data and calculate any derivations in the component itself.

src/types/paths.ts
```diff
@@ -0,0 +1,88 @@
...
+      ? GetValueAtPath<T[P], Rest>
+      : never
+    : Path extends keyof T
+    ? T[Path]
+    : never
+  : never;
+
+export type PathsAndValueType<T> = {
+  [K in Paths<T> as K extends string ? K : never]: GetValueAtPath<T, K>;
+};
```
sergei-maertens, line 84: unused -> drop it?

src/types/paths.ts
sergei-maertens: I had to [play around](https://www.typescriptlang.org/play?ts=4.9.5#code/C4TwDgpgBAogHgYwDYFcAmEBCBDAThAeQCMArCBYAHgBUA+KAXimqgjmAgDs0BnKAJXIB7XGko9guAJacA5gBooKTgGtOQgO6daAKChQA-M1bsuvKADFlFKUM579hqJwgA3CLgf6AXMwe+Xd1wAbh0HUEgoAAVsYAALHhp6JhY2Dm4+agcjAG8vRwBtAGkoGSgVCBAhADNmAF0AWgNfErSzPglpOXzHJ3hkdCw8QlJyKmpiuvo2jKghUYoe3qcSgB8oAAMAEhyigF8AOh2Y+MT+1AwcfGIyChpJ2lo9jaXelteoALcPUN69goqVVq1Dq-mc3xCYX0EWgAHEIMAAGrYVAQACCwBOcRoiixJnS5k6MlkyWMM3MWX0Rjx5L42xyMmqHmihx2jOZggkz3y1Px7XKlRqfmWTnhSJRKHRmNi2ImUTqik5wF0Iq+QXyvhpplmgKFlMcRjlMtBb3B6p8Zp+YRh0RlPDR3GRqOo4AgSUYUDy+mKpU4ttO7uwfFa2sJkmJK0+ltwdV8YqdkoxWJxUCKtFCe1COhtWJ4FhECYgLsgKcRpN1tS9UAKUV9-oSgb48YlUuT1Fx01DfERTlrao8sagkklGazOhkHFw1WwCGgBDAwFsfqr+lcLd8RLkv0cSGwRAgSA34a3Oj21tdUAAykIALaEBdLm5jD0r6tFI9dWSD+eLuwFOqjuekQWB6uaUNed4-o+CzKqECB2BIUDVL4IFMAA5EQu6YdgAD02CcCA8TEmhWY2gQJRMLmDpoIWxZuhB96-pwT4UOmQHQNQECIZRdr5rgtGuuBt6MdBtzAIoUF-lMcEIcAQ5ccAvicdxUBoSR7FQAAipKHoVp6+Q1nWIFBlAzaokmMpCZBD52Cx4nRJ2BJ8JJnD-r2UaBAOvjDhAp46DhOFQDwQhQBoUjxEFx6yFAJkyBgcBDheMgSBA2BoHMtTYFASDhR4KJKJwS6KBocSxKFQgoEg6XxNA+A3tgYCQOlgLZUIQgqNlUgVAY2YXtpEAAEwemZibSvEVkibZMGKJuJJ8rMLluUYs0eRCoQBaUclIG1Kh8DlFSRZ+ACEvWRP1ADMumCpWxQfsSg4+ZmUAbSt6ycCgN77p4OEAFQ-XoP1QGiAASaJHUOcRSHwM7ACgKJICAUD1RUHRcDwECKPuCDYCg6NQFUKBQNjfoExD+DQDC0M8ADcwwb6aBSLOPB3XIijvZ9zL4elPAgJ9QhIAcV680Q-OlHwGgQFIogxdwRP4VA+6lLI6j4Gg8g0+FaHQ0gwVIdgUsI+CGgy+l6hyWAu6yJK6VEIjwNQ8AIiM-lmCXgcNM08QBTs19dQxRQcNIIb8GcEEwB8I7MWHRGE4eJw8MgIod74RG8MCiA1NQIDwzR3IPCC9eEPQDIQR4w1FtSFxQ5tVADShXEXD4xVUCle4MsxbgRDhbgeCI-MYnq1nTeE8TaFyWomhISIQUoPgENlTVvqTvHSC59FLVQ84QhyVlPseDNQg09U0+L3vuCJZE4VQGAuBCGgKBM2vUBvR9X0A4F-mBQAAuHDRsJAFA-64Fvp4EOiFj4hSYDkAAjL4aBA1zqKDQtAtCvg0J7gQGhTMOgwH8wgAcbasgAAUECACUWYNolTKlIWoEskZCDbovEarYZTVwEMDS8PUbRFAgB1JgelXyGTKBYQczCLLjQYi5OyHZ5rmEWn7Xk-ZPBnn8n9Gm1BcCI0jtUKQCUiJ5wVojba7UIwoDAEXKAa5UR1kXvwDhvoUppQyunKAdVy7Eg9kPA43it4xTXFIJAihF4bw6IuIORc55cwsXYy8YtozFUhggOIctOBjyChACAN5q400VjopAk4IDpQqnJbwRD34aUwLfConBcx8QEiWdsUAywvgcEIv0YE6CiIRIWcRsoZG0iae5JRgFxycEnNOWcV4DxjEvEkzJWVXwwjQejJAYx1K9DQLEbAl5cAIDQVYyUPB1mOE2cAbAvhXy9AOVxXw8jtz6DPCo2OU4ZzQH4GlWwsyG71X0r0JZqke4MyEMc1cLZmZQDuX5U60BLzTIoLmD0lT2pcFqQWFsdFwJwuAF8+ZEkbKuWkjg2S6TVnwplL4PSsLSVjQSB6NCpzsAHGuUcmSnBEIrLGFiAaFLrpTOpQi9CDKdmYOCPoDapSgpxAqlVFu2A277kbnkgpRSUDADCGAuSGgRAqGJFSzl5L6x5jRc6QSeqKA4vqnipi-5ST0q2UysF6kNoeBAVTEwACOBoCJWyzV2rdVYq5ZqXixrJQYrNdiuZlqIX4ptXSoVuySJAA) a bit to figure out why this produces that weird union type.

In short - when looking at a plain JS object with unspecified literal keys, `keyof Object` is `string | number` since both are valid to index the object. Number is converted to string, since internally only strings can be used to index.

`keyof {[K: string]: any} == string | number`

When order is respected in an object, try running this in the browser:

```js
console.log({123: 1, '123': 'a'});
// 'a'
```

Viicos: Ok makes sense now, another reason JS is such a broken language

src/types/paths.ts
```diff
@@ -0,0 +1,88 @@
+// These can cause infinite recursions, and since we're dealing with JSON-serializable
+// data only, the keys of the objects must be strings.
+type ExcludeBareObject<T> = T extends Record<string, unknown>
+  ? T extends Function
+    ? never
+    : T
+  : never;
+
+/**
+ * Constructs a union of possible paths into a given object type.
```
Viicos, line 10:
```suggestion
 * Constructs a union of possible paths as dotted strings into a given object type.
```
Or similar?

Viicos (approved):

src/types/paths.ts
Viicos: Maybe keep it but add a comment saying it's related but not used?

src/types/paths.ts
sergei-maertens: :thisisfine:

src/types/paths.ts
sergei-maertens: nope, we can easily re-add it or find it via this PR if we need it. Let's not clutter our codebases with dead/unused code, keeping the used code in check is already enough of a challenge.

PR #50, opened 2023-11-06
✨ Add `currency` component

Viicos:


src/registry/currency/edit-validation.ts
```diff
@@ -0,0 +1,36 @@
...
+const singleValueSchema = z
+  .object({multiple: z.literal(false)})
+  .and(z.object({defaultValue: currencySchema}));
+
+// case for when component.multiple=true
+const multipleValueSchema = z
+  .object({multiple: z.literal(true)})
+  .and(z.object({defaultValue: currencySchema.array()}));
+
+const defaultValueSchema = singleValueSchema.or(multipleValueSchema);
```
Viicos, line 21: Should currency also be only single capable, like number?

src/registry/currency/edit.tsx
```diff
@@ -0,0 +1,214 @@
...
+      label={
+        <FormattedMessage
+          description="Label for 'allowNegative' builder field"
+          defaultMessage="Allow negative values"
+        />
+      }
+      tooltip={tooltip}
+    />
+  );
+};
```
Viicos, line 212: Should we move this to `components/builder`?

src/registry/currency/edit.tsx
```diff
@@ -0,0 +1,214 @@
...
+  const tooltip = intl.formatMessage({
+    description: "Tooltip for 'defaultValue' builder field",
+    defaultMessage: 'This will be the initial value for this field before user interaction.',
+  });
+  return (
+    <NumberField
+      name="defaultValue"
+      label={<FormattedMessage {...LABELS.defaultValue} />}
+      tooltip={tooltip}
+    />
```
Viicos, line 170: In the current form builder, as soon as a value is being entered by the user, it is prefixed by `EUR`. Should we replicate the same here? If so, I was thinking about adding a new prop to the `NumberField` component for that. I'm not sure how to call it though, `prefix` feels wrong as formio renders it like the following:
image

In our case, it more or less a "placeholder prefix"

src/registry/currency/edit-validation.ts
sergei-maertens: yeah let's drop the "multiple" behaviour. I can't make sense of it.

It would be wise to write a detection script in the backend so that we check if any (prod) instances make use of it.

src/registry/currency/edit-validation.ts
```diff
@@ -0,0 +1,36 @@
...
+
+const currencySpecific = z.object({
+  decimalLimit: z.number().int().positive().optional(),
```
sergei-maertens, line 24: the types restrict this to the values 0-9 (maybe 10), so perhaps we should do the same here?

`z.union([z.literal(0), z.literal(1), z.literal(2)])` etc. (you can write a nice `Array(9).map` for this probably).

src/registry/currency/edit.tsx
```diff
@@ -0,0 +1,213 @@
...
+  showInSummary: true,
+  showInEmail: false,
+  showInPDF: true,
+  hidden: false,
+  clearOnHide: true,
+  isSensitiveData: false,
+  defaultValue: undefined,
+  decimalLimit: undefined,
+  allowNegative: false,
+  currency: 'EUR',
```
sergei-maertens, line 137: I like to order the fixed/hardcoded values to above the `// basic tab` comment, they don't really belong to a particular group in the edit form

sergei-maertens, line 137: I like to order the fixed/hardcoded values to above the `// basic tab` comment, they don't really belong to a particular group in the edit form

src/registry/currency/edit.tsx
sergei-maertens: I think prefix would be perfect for this, we render it slightly nicer in our own frontend (check the number component suffix stories) and you have accessibility gains because you can (and should) include the prefix in the label, so a screen reader would read out "the amount to refund, EUR <input value>". That gives screen reader users a good hint about which currency is expected.

If this is only injected once a value is entered, this information is too late and we can't properly include it in the accessible label.

Using the actual euro sign instead of EUR would probably be the best here too.

src/registry/currency/edit.tsx
sergei-maertens: let's not yet - they are not generic, just happen to be used in both currency/number component but those are basically parent & child. AllowNegative is scheduled for removal too.

src/registry/currency/edit-validation.ts
Viicos: Had to cast `as any` because it doesn't typecheck somehow

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #51, opened 2023-11-06
✨ Add `iban` and `licenseplate` component

Viicos:


Viicos:
The type definition of `licenseplate` seems to be incorrect:

image

```ts
export interface LicensePlateProperties {
    type: 'licenseplate';
    validate: {
        pattern: '^[a-zA-Z0-9]{1,3}\\-[a-zA-Z0-9]{1,3}\\-[a-zA-Z0-9]{1,3}$';  // Should also extend OFValidateOptions?
    };
    validateOn: 'blur';
}
```

sergei-maertens:
> The type definition of `licenseplate` seems to be incorrect:
> 
> image
> 
> ```ts
> export interface LicensePlateProperties {
>     type: 'licenseplate';
>     validate: {
>         pattern: '^[a-zA-Z0-9]{1,3}\\-[a-zA-Z0-9]{1,3}\\-[a-zA-Z0-9]{1,3}$';  // Should also extend OFValidateOptions?
>     };
>     validateOn: 'blur';
> }
> ```
> 
> Also, I've removed readonly input checkboxes in last commit for these two components, but I don't see any reference to that attribute in the types, nor in the base formio types?

Good catch, something seems to have gone wrong there. We should at least check the type tests to see if those also highlight the issue. There may be a forgotten intersection or something like that.

Viicos:
Hum, there's [a test for this](https://github.com/open-formulieren/types/blob/17777fa93720bdc893c548d88be75a412fb30a4a/test-d/formio/components/licenseplate.test-d.ts#L46-L76) and it seems to pass

sergei-maertens:
> Hum, there's [a test for this](https://github.com/open-formulieren/types/blob/17777fa93720bdc893c548d88be75a412fb30a4a/test-d/formio/components/licenseplate.test-d.ts#L46-L76) and it seems to pass

interesting :thinking:

src/registry/licenseplate/preview.tsx
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * Show a formio iban component preview.
+ *
+ * NOTE: for the time being, this is rendered in the default Formio bootstrap style,
+ * however at some point this should use the components of
+ * @open-formulieren/formio-renderer instead for a more accurate preview.
+ */
+const Preview: React.FC<ComponentPreviewProps<LicensePlateComponentSchema>> = ({component}) => {
+  const {key, label, description, tooltip, validate = {}, multiple} = component;
+  const {required = false} = validate;
```
sergei-maertens, line 16: isn't it because of the `validate = {}` default? because `component.validate` itself can be undefined.

Viicos: For some reason, using `{}` as a default value doesn't seem to play well if `validate` is overridden in the types library (either as `validate` or `validate?`):
```ts
const {key, label, description, tooltip, validate = {}, multiple} = component;  // `validate` is of type `{}`
const {key, label, description, tooltip, validate, multiple} = component;  // `validate` is of type OFValidateOptions<Validator | "plugins"> & { "^[a-zA-Z0-9]{1,3}\\-[a-zA-Z0-9]{1,3}\\-[a-zA-Z0-9]{1,3}$";}
```

Fortunately types define `validate` as being required, so we can safely remove the default value.

May be related: https://github.com/microsoft/TypeScript/issues/41189

sergei-maertens: Yeah, perhaps this should be `= ({} satisfies OFValidate<...>)` in this case, right now it's probably interpreted as a literal.

Doing the destructuring differently here is acceptable for me :+1:

src/registry/iban/edit.tsx
```diff
@@ -0,0 +1,181 @@
...
+            'showInSummary',
+            'showInEmail',
+            'showInPDF',
+            'multiple',
+            'hidden',
+            'clearOnHide',
+            'isSensitiveData',
+            'defaultValue',
+            'disabled',
+            'placeholder'
```
sergei-maertens, line 66: these are not exposed in the builder (type checking of these will come soon!)

src/registry/iban/edit.tsx
```diff
@@ -0,0 +1,181 @@
...
+  tooltip: '',
+  showInSummary: true,
+  showInEmail: false,
+  showInPDF: true,
+  multiple: false,
+  hidden: false,
+  clearOnHide: true,
+  isSensitiveData: false,
+  defaultValue: '',
+  placeholder: '',
```
sergei-maertens, line 142: I'm surprised the type checker didn't complain about this one though

src/registry/licenseplate/edit.tsx
```diff
@@ -0,0 +1,180 @@
...
+            'tooltip',
+            'showInSummary',
+            'showInEmail',
+            'showInPDF',
+            'multiple',
+            'hidden',
+            'clearOnHide',
+            'isSensitiveData',
+            'defaultValue',
+            'disabled'
```
sergei-maertens, line 65:
```suggestion
```

src/registry/iban/edit.tsx
```diff
@@ -0,0 +1,181 @@
...
+  key: '',
+  description: '',
+  tooltip: '',
+  showInSummary: true,
+  showInEmail: false,
+  showInPDF: true,
+  multiple: false,
+  hidden: false,
+  clearOnHide: true,
+  isSensitiveData: false,
```
sergei-maertens, line 140:
```suggestion
  isSensitiveData: true,
```
bank account numbers are (privacy) sensitive :)

src/registry/licenseplate/edit.tsx
```diff
@@ -0,0 +1,180 @@
...
+  key: '',
+  description: '',
+  tooltip: '',
+  showInSummary: true,
+  showInEmail: false,
+  showInPDF: true,
+  multiple: false,
+  hidden: false,
+  clearOnHide: true,
+  isSensitiveData: false,
```
sergei-maertens, line 139:
```suggestion
  isSensitiveData: true,
```
possibly personally identifiable information -> mark as sensitive data by default

sergei-maertens (changes requested):
Spotted a couple of oversights.

src/registry/iban/edit.tsx
Viicos: It seems types are incorrect: `defaultValue` is not excluded. But Iban is multiple capable, so according to the definition of `MultipleCapable`, it would infer `never` if we `Omit` `defaultValue` on the `BaseIbanComponentSchema` :/

sergei-maertens (approved):

Issue #35, opened 2023-09-04
Properly implement tooltips


sergei-maertens:
The current tooltip solution is quick and dirty and has some issues:

* https://github.com/open-formulieren/formio-builder/pull/33#discussion_r1314738060 z-index fighting
* Long texts fall of the screen or sit behind other panels in multi-column layouts
* Hovering over them doesn't really allow clicking any potential URLs inside or copy-pasting something. Not necessarily a dealbreaker, but could be annoying
* Hover on desktop displays the tooltip but *also* the `title` HTML attribute.

A proper implementation would be to use https://floating-ui.com/docs/react, which we already use in the SDK for date pickers. It would also solve some other issues in the SDK/renderer eventually by not having to use the tooltip library of formio itself.

PR #53, opened 2023-11-15
Implement the builder form for the selectboxes component

### LINKED ISSUE #52: Implement select component edit form

sergei-maertens:
Closes #52

Ho boi this got a lot bigger than I expected...

### DIFF
--- a/src/components/ComponentConfiguration.stories.tsx
+++ b/src/components/ComponentConfiguration.stories.tsx
@@ -777,6 +777,10 @@ export const FileUpload: Story = {
     component: {
       id: 'kiweljhr',
       storage: 'url',
+      webcam: false,
+      options: {
+        withCredentials: true,
+      },
       url: '',
       type: 'file',
       key: 'file',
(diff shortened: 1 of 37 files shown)


src/components/ComponentConfiguration.stories.tsx
```diff
@@ -900,3 +904,35 @@ export const FileUpload: Story = {
...
+      icon: 'plus-square',
+      group: 'basic',
+      weight: 60,
+      schema: {},
+    },
+  },
+
+  play: async ({canvasElement, step, args}) => {
+    const canvas = within(canvasElement);
+  },
```
sergei-maertens, line 937: obviously TODO :)

src/components/builder/values/values-config.tsx
```diff
@@ -0,0 +1,37 @@
...
+ */
+export function ValuesConfig<T extends SchemaWithDataSrc>({name}: ValuesConfigProps<T>) {
+  const {
+    values: {
+      openForms: {dataSrc},
+    },
+  } = useFormikContext<T>();
+  return (
+    <>
+      <ValuesSrc />
```
sergei-maertens, line 30: best would be to pass an `onChange` here so that we can reset the appropriate field

src/registry/selectboxes/edit.tsx
```diff
@@ -0,0 +1,183 @@
...
+        <Key isManuallySetRef={isKeyManuallySetRef} generatedValue={generatedKey} />
+        <Description />
+        <Tooltip />
+        <PresentationConfig />
+        <Hidden />
+        <ClearOnHide />
+        <IsSensitiveData />
+        <ValuesConfig<SelectboxesComponentSchema> name="values" />
+        <DefaultValue options={options} />
+        {/*<DefaultValue multiple={!!values.multiple} />*/}
```
sergei-maertens, line 92: TODO

src/registry/selectboxes/preview.tsx
```diff
@@ -0,0 +1,56 @@
...
+          <div key={`option-${value}-${index}`} className="form-check">
+            <label className="form-check-label">
+              <CheckboxInput name={`${key}.${value}`} label={label} />
+            </label>
+          </div>
+        ))}
+      </div>
+
+      {description && <Description text={description} />}
+    </Component>
```
sergei-maertens, line 52: this will likely be required for the defaultValue anyway

src/components/builder/values/values-config.stories.tsx
```diff
@@ -0,0 +1,201 @@
...
+
+    const doSubmit = async () => {
+      const btn = canvas.getByRole('button', {name: 'Submit'});
+      // https://github.com/testing-library/user-event/issues/1075
+      // apparently userEvent.click(submitButton) doesn't do anything in Firefox :/
+      fireEvent.click(btn);
+      await waitFor(async () => {
+        await expect(canvas.queryByText('Submitting...')).toBeNull();
+      });
+    };
```
sergei-maertens, line 121: I was under the impression that testing was quite nice these days with testing-library, but this issue and realization really made me lose confidence again :/

src/components/builder/values/values-config.tsx
sergei-maertens: I tried that and it leads to race conditions, so I'm now using `useLayoutEffect` instead.

There might be an option to use React 18's `flushSync` instead, but the builder & backend are still stuck on React 17 currently :(

src/registry/selectboxes/edit-validation.ts
sergei-maertens, line 1: Need to test this still & add interaction tests

src/registry/selectboxes/edit.tsx
```diff
@@ -0,0 +1,229 @@
...
+    // values for options that were removed.
+    const explicitDefaults: SelectboxesComponentSchema['defaultValue'] = {};
+    optionValues.forEach(optionValue => {
+      // if a value is specified already in the form state, use it, otherwise default to "unchecked".
+      const defaultForOption = value.hasOwnProperty(optionValue) ? value[optionValue] : false;
+      explicitDefaults[optionValue] = defaultForOption;
+    });
+
+    setFieldValue('defaultValue', explicitDefaults);
+    console.log('normalized defaultValue');
```
sergei-maertens, line 216: TODO: remove

src/components/builder/values/items-expression.tsx
```diff
@@ -0,0 +1,51 @@
...
+    >
+      <div>
+        <JSONEdit name={NAME} data={value} rows={3} id={htmlId} />
+      </div>
+
+      <Description
+        text={
+          <FormattedMessage
+            description="Description for the 'openForms.itemsExpression' builder field"
+            defaultMessage={`A JSON logWic expression returning a variable (of array type)
```
Viicos, line 42:
```suggestion
            defaultMessage={`A JSON Logic expression returning a variable (of array type)
```
?

src/components/builder/values/values-table.tsx
SilviaAmAm, line 1: I thought we had validation for invalid characters in the keys of the options, but I checked our test env and formio builder and it is not the case. I would have sworn that we had had problems with dots and spaces :thinking: 

Another thing is that we don't validate for unique keys by the options. So if you have multiple options with  different labels but the same options, then as you click one also the other one is selected

src/components/formio/textarea.tsx
```diff
@@ -0,0 +1,97 @@
...
+  description = '',
+  showCharCount = false,
+  onChange,
+  ...props
+}) => {
+  const {getFieldProps, getFieldMeta} = useFormikContext();
+  const {value, onChange: formikOnChange} = getFieldProps<string | undefined>(name);
+  const {touched} = getFieldMeta<string | undefined>(name);
+  const {errors, hasErrors} = useValidationErrors(name);
+  // const [{value}, {touched}] = useField<string | undefined>(name);
```
SilviaAmAm, line 37: Is this left here as todo?

src/components/formio/checkbox.tsx
```diff
@@ -23,32 +50,16 @@ const Checkbox: React.FC<CheckboxProps> = ({
...
-    </Component>
-  );
-};
-
+}) => (
+  <Component field={name} required={required} type="checkbox">
+    <div className="form-check checkbox">
+      <label className="form-check-label">
+        <CheckboxInput name={name} label={label} onChange={onChange} />
+        {tooltip && ' '}
```
Viicos, line 58: I don't know much on this side, is there a better way of adding spacing between these two elements?

src/registry/selectboxes/helpers.ts
```diff
@@ -0,0 +1,10 @@
+import {SelectboxesComponentSchema} from '@open-formulieren/types';
+import {Option} from '@open-formulieren/types/lib/formio/common';
+
+// A type guard is needed because TS cannot figure out it's a discriminated union
+// when the discriminator is nested.
```
Viicos, line 5: Maybe reference the TS issue so that we can track it more easily? https://github.com/microsoft/TypeScript/issues/18758

src/registry/validation.ts
```diff
@@ -74,6 +74,46 @@ export const buildCommonSchema = (intl: IntlShape) =>
...
+          description: 'Form builder option label required error',
+          defaultMessage: 'The option label is a required field.',
+        }),
+      })
+      .min(1),
+    openForms: z
+      .object({
+        // zod doesn't seem to be able to use our supportedLanguageCodes for z.object keys,
+        // they need to be defined statically. So, 'record' it is.
+        translations: z.record(optionTranslationSchema.optional()),
```
Viicos, line 112: Does it work when specifying `z.record` keys as `z.union([z.literal('en'), z.literal('nl')])`? (Or built dynamically from `supportedLanguageCodes`, although you might need to ts ignore or use `as smth`)

Viicos (approved):

src/components/builder/values/values-table.tsx
```diff
@@ -31,19 +31,37 @@ export interface ValuesTableProps<S> {
...
+  const intl = useIntl();
   const {getFieldProps} = useFormikContext();
   const {value: options = []} = getFieldProps<Option[] | undefined>(name);
 
+  const tooltip = intl.formatMessage({
+    description: "Tooltip for 'values' builder field",
+    defaultMessage: `The values that can be picked for this field. Values are the text
+    that is submitted with the form data. Labels are the text next to radio buttons,
+    checkboxes and options in dropdowns.
+    `,
```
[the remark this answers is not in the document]
sergei-maertens: The tooltips not behaving ideally is a known issue, it also overflows in other places. We need to fix the tooltip library rather than the content inside the tooltip. See #35.

Adding the param for the component name can be a nice future improvement, let's maybe see how the re-design turns out for that first?

src/components/formio/checkbox.tsx
sergei-maertens: Unfortunately not, React collapses all whitespace and if they're different components, it even removes the whitespace.

src/registry/selectboxes/helpers.ts
sergei-maertens: Oh I didn't actually find the TS issue but just concluded this from various TS playground experiments :P I will add it

src/components/builder/values/values-table.tsx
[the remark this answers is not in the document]
Viicos: > Adding the param for the component name can be a nice future improvement, let's maybe see how the re-design turns out for that first?

Yes this is really a small detail anyway

src/components/builder/values/values-table.tsx
sergei-maertens: I thought so too, perhaps that functionality got lost when we added the `itemsExpression` functionality, since that touched the builder/edit form configuration?

Hooking up the key validation should be easy, hooking up the unique values should also be reasonable so I'll get started on those.

src/components/formio/textarea.tsx
sergei-maertens: nope, it's probably left over debugger stuff. I'll remove it.

src/registry/validation.ts
sergei-maertens: I tried building it dynamically and that doesn't work, and specifying them literally here defeats the purpose of having the `supportedLanguageCode`s in the builder context

src/components/builder/values/values-table.tsx
sergei-maertens: Discussed in the office and the result is a separate issue: #54 - the behaviour in this PR is consistent with the current (and upstream Formio) behaviour.

src/registry/validation.ts
Viicos: I've tried:

```ts
const supportedLanguageCodes = ['en', 'nl'];
z.record(z.union(supportedLanguageCodes.map(z.literal)), optionTranslationSchema.optional())
```

and it seems to work (but with a fixed value of `supportedLanguageCodes`, it's not coming from the context or anything). Am I missing something?

sergei-maertens: the fixed value of `supportedLanguageCodes` is the problem - we centralize the list of supported language codes so that we don't have to update code in multiple places when we add additional languages.

And as you saw in the storybook build, that is set up with just a single language code, which causes `zod` to break because the `union` requires at least 2 elements in the union array.

Viicos: > And as you saw in the storybook build, that is set up with just a single language code, which causes `zod` to break because the `union` requires at least 2 elements in the union array.

This is what I was missing :+1:

PR #57, opened 2023-11-20
Implement radio component edit form

### LINKED ISSUE #56: Implement radio component edit form
Related to open-formulieren/open-forms#2958

sergei-maertens:
Closes #56

Very similar to the selectboxes component, so the code is mostly copy pasted :grimacing:

### DIFF
--- a/src/components/formio/radio.stories.ts
+++ b/src/components/formio/radio.stories.ts
@@ -1,0 +1,87 @@
+import {expect} from '@storybook/jest';
+import {Meta, StoryObj} from '@storybook/react';
+import {within} from '@storybook/testing-library';
+
+import {withFormik} from '@/sb-decorators';
+
+import Radio from './radio';
+
+export default {
+  title: 'Formio/Components/Radio',
+  component: Radio,
+  decorators: [withFormik],
+  parameters: {
+    modal: {noModal: true},
+    formik: {initialValues: {'my-radio': ''}},
+  },
+  args: {
+    name: 'my-radio',
+    label: '',
+    tooltip: '',
+    options: [
+      {value: 'a', label: 'A'},
+      {value: 'b', label: 'B'},
+    ],
+  },
+} as Meta<typeof Radio>;
+
+type Story = StoryObj<typeof Radio>;
+
+export const Default: Story = {
+  args: {
+    label: 'Labeled radio',
+  },
+};
+
+export const WithInitialChecked: Story = {
+  parameters: {
+    formik: {
+      initialValues: {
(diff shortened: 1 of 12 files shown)


src/components/formio/radio.stories.ts
```diff
@@ -0,0 +1,87 @@
...
+      {value: 'b', label: 'B'},
+    ],
+  },
+} as Meta<typeof Radio>;
+
+type Story = StoryObj<typeof Radio>;
+
+export const Default: Story = {
+  args: {
+    label: 'Labeled selectboxes',
```
Viicos, line 32:
```suggestion
    label: 'Labeled radio',
```

src/components/formio/radio.stories.ts
```diff
@@ -0,0 +1,87 @@
...
+export const WithInitialChecked: Story = {
+  parameters: {
+    formik: {
+      initialValues: {
+        'my-radio': 'b',
+      },
+    },
+  },
+  args: {
+    label: 'Labeled selectboxes',
```
Viicos, line 45:
```suggestion
    label: 'Labeled radio',
```

src/registry/radio/helpers.ts
```diff
@@ -0,0 +1,11 @@
...
+import {Option} from '@open-formulieren/types/lib/formio/common';
+
+// A type guard is needed because TS cannot figure out it's a discriminated union
+// when the discriminator is nested.
+// See https://github.com/microsoft/TypeScript/issues/18758
+export const checkIsManualOptions = (
+  component: RadioComponentSchema
+): component is RadioComponentSchema & {values: Option[] | undefined} => {
+  return component.openForms.dataSrc === 'manual';
+};
```
Viicos, line 11: Can't we have this generic? Something along the lines of

```suggestion
export const checkIsManualOptions<T> = (
  component: T
): component is T & {values: Option[] | undefined} => {
  return component.openForms.dataSrc === 'manual';
};
```

Viicos (changes requested):

src/registry/radio/helpers.ts
sergei-maertens: no because `select` used a different shape for the `values`, it's nested inside `data`. So I consider moving this further away from where it's used and adding generics too much indirection and a little bit of duplicated code like this is totally acceptable to me, but if you have strong feelings about it I'm willing to reconsider

Viicos: Ok now makes sense if structure is different, better to have duplicated code rather than hard to understand generics :)

PR #58, opened 2023-11-21
Add `textarea` component

Viicos:


src/registry/textarea/edit-validation.ts
```diff
@@ -0,0 +1,29 @@
...
+// case for when component.multiple=true
+const multipleValueSchema = z
+  .object({multiple: z.literal(true)})
+  .and(z.object({defaultValue: textareaSchema.array()}));
+
+const textareaSpecific = z.object({
+  validate: z.object({
+    maxLength: z.number(),
+  }),
+});
```
Viicos, line 22: Am I missing anything? Should I also add textarea specific fields?

src/registry/textarea/edit.tsx
```diff
@@ -0,0 +1,243 @@
...
+        <Multiple<TextareaComponentSchema> />
+        <Hidden />
+        <ClearOnHide />
+        <IsSensitiveData />
+        <DefaultValue multiple={!!values.multiple} />
+        <AutoComplete />
+        <ReadOnly />
+        <Placeholder />
+        <ShowCharCount />
+        <AutoExpand />
```
Viicos, line 101: I don't know if `AutoExpand` could be reflected in the preview? It doesn't seem to be supported in plain HTML

src/registry/textarea/edit-validation.ts
sergei-maertens: many components don't necessarily have specific validations, I think this one is even overkill because the max length validation input should only allow numbers in the first place.

*if* we go the validation route here (which adds extra robustness), it should be as specific as possible:

```tsx
z.number().int().gte(1).optional()
```
to ensure that we don't get floats, a value is at least 1 if provided and the `.optional()` to allow it to be unspecified.

src/registry/textarea/edit.tsx
sergei-maertens: Looking at https://stackoverflow.com/a/65990608 it doesn't seem to be a lot of work, could easily be implemented as a generic hook even.

W/r to this builder it's a bit of a tradeoff - if it's not a lot of work, yep, let's get the behaviour implemented (the added benefit is that it informs us for the future implementation in the renderer), if it's too much work then we might get by with making it appear similarly (e.g. the `inputMask` functionality is mimicked via `placeholder` attribute, since it's waaaay too much work and has accessibility issues too).

Viicos: The SO answer worked fine

src/components/formio/textarea.tsx
```diff
@@ -37,6 +39,13 @@ export const TextArea: React.FC<JSX.IntrinsicElements['textarea'] & TextAreaProp
   const inputRef = useRef<HTMLInputElement>(null);
   const {bareInput} = useContext(RenderContext);
 
+  useLayoutEffect(() => {
+    if (autoExpand && inputRef.current) {
+      inputRef.current.style.height = 'inherit';
+      inputRef.current.style.height = `${inputRef.current.scrollHeight}px`;
+    }
+  }, [value]);
```
sergei-maertens, line 47:
```suggestion
  }, [autoExpand, value]);
```
I still need to look into setting up eslint in this project, but if `autoExpand` toggles between true/falls then this hook also needs to re-evaluate :)

Additionally, I think that `inputRef` should also be passed to the dependencies (but it has a stable identity).

src/registry/textarea/edit-validation.ts
```diff
@@ -0,0 +1,31 @@
...
+const multipleValueSchema = z
+  .object({multiple: z.literal(true)})
+  .and(z.object({defaultValue: textareaSchema.array()}));
+
+// Omit `autoExpand` as it will be set in the UI via a checkbox
+const textareaSpecific = z.object({
+  validate: z.object({
+    maxLength: z.number().int().gte(1).optional(),
+  }),
+  rows: z.number().int().gte(1),
```
sergei-maertens, line 23: this can also be optional, no?

src/registry/textarea/preview.tsx
```diff
@@ -0,0 +1,48 @@
...
+    label,
+    description,
+    placeholder,
+    tooltip,
+    validate = {},
+    autocomplete = '',
+    disabled = false,
+    multiple,
+    showCharCount,
+    rows = 3,
```
sergei-maertens, line 26: I wouldn't give this a default value since it can be left empty, no?

src/registry/textarea/edit.tsx
```diff
@@ -0,0 +1,245 @@
...
+    />
+  );
+};
+
+const AutoExpand: React.FC<{}> = () => {
+  const intl = useIntl();
+  const tooltip = intl.formatMessage({
+    description: "Tooltip for 'AutoExpand' builder field",
+    defaultMessage:
+      "This will make the text area auto expand it's height as the user is typing into the area.",
```
sergei-maertens, line 209:
```suggestion
      "This will make the text area auto expand its height as the user is typing into the area.",
```

src/registry/textarea/edit.tsx
```diff
@@ -0,0 +1,245 @@
...
+  },
+  translatedErrors: {},
+  registration: {
+    attribute: '',
+  },
+};
+
+interface DefaultValueProps {
+  multiple: boolean;
+  rows: number;
```
sergei-maertens, line 184: I think this can be ignored and we can pin it to 3 rows in the builder edit form.

src/registry/textarea/edit-validation.ts
Viicos: It is marked as required in types :shrug:

src/components/formio/textarea.tsx
Viicos: > Additionally, I think that `inputRef` should also be passed to the dependencies (but it has a stable identity).

Isn't using the value sufficient? What changes on `inputRef` could require the hook to be triggered?

src/registry/textarea/preview.tsx
Viicos: Types also mark it as required, but I think it shouldn't be in this case?

src/registry/textarea/edit-validation.ts
sergei-maertens: Yeah I see formio itself sets a default, but also supports falsy values: https://github.com/formio/formio.js/blob/4.13.x/src/components/textarea/TextArea.js#L53

I think we should make it optional in the types indeed.

src/registry/textarea/preview.tsx
sergei-maertens: especially if the type marks it required I wouldn't expect a default value in the destructuring, but following https://github.com/open-formulieren/formio-builder/pull/58#discussion_r1400788984 we should be able to just pass it along whether a value is set or not (undefined) without a default value.

src/components/formio/textarea.tsx
sergei-maertens: it's definitely something that eslint will complain about - and technically it *can* change even though it's a ref, consider this contrived example:

```tsx
const [value, setValue] = useState('foo');
const ref1 = useRef(null);
const ref2 = useRef(null);

let hookRef = ref1;
if (value === 'bar') {
  hookRef = ref2;
}

useLayoutEffect(
  () => {...},
  [hookRef],
);
```

So you can have the 'target' ref change for a particular hook depending on some other state or props.

This is why eslint essentially enforces that anything that is inside your hook body should be in the dependency array. It's also robustness for future changes in your code - ignoring the linter rule is dangerous since future requirements may require changes to the body of the hook and then it's easy to forget to add any relevant new dependencies.

In this case, the ref indeed does not change, so it will *also* not trigger a new evaluation of the hook body, meaning the end result is the same and you'll be able to satisfy the linter.

src/registry/textarea/preview.tsx
Viicos: https://github.com/open-formulieren/types/pull/28

sergei-maertens: Merged & release tag pushed, it should be on npm in a couple of minutes

sergei-maertens (approved):

PR #60, opened 2023-11-21
Implement the select component edit form

### LINKED ISSUE #59: Implement the select component in the builder
Should be pretty much a copy of the radio component, but with slightly different component schema and of course a dropdown instead of radio items.

sergei-maertens:
Closes #59

### DIFF
--- a/src/components/builder/values/values-config.tsx
+++ b/src/components/builder/values/values-config.tsx
@@ -8,6 +8,14 @@ import {SchemaWithDataSrc} from './types';
 import ValuesSrc from './values-src';
 import ValuesTable, {ValuesTableProps} from './values-table';
 
+/**
+ * Check if the dotted `path` exists on `obj`.
+ *
+ * @example
+ * ```
+ * isNestedKeySet({my: {path: 'irrelevant'}}, 'my.path') // true
+ * ```
+ */
 function isNestedKeySet(obj: {}, path: string): boolean {
   const bits = path.split('.');
   for (const bit of bits) {
--- a/src/components/builder/values/values-config.tsx
+++ b/src/components/builder/values/values-config.tsx
@@ -1,10 +1,25 @@
 import {useFormikContext} from 'formik';
 import {useLayoutEffect} from 'react';
 
+import {hasOwnProperty} from '@/types';
+
 import ItemsExpression from './items-expression';
 import {SchemaWithDataSrc} from './types';
 import ValuesSrc from './values-src';
 import ValuesTable, {ValuesTableProps} from './values-table';
+
+function isNestedKeySet(obj: {}, path: string): boolean {
+  const bits = path.split('.');
+  for (const bit of bits) {
+    // as soon as any node does not have the respective path set, exit, the full deep
+    // path will then also not be set.
+    if (!hasOwnProperty(obj, bit)) {
+      return false;
+    }
+    obj = obj[bit] as {};
+  }
+  return true;
+}
 
 export interface ValuesConfigProps<T> {
   name: ValuesTableProps<T>['name'];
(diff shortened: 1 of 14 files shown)


Viicos:
I am experiencing a weird behavior with datasource:
- a `*` appears in the select when "From variable" is selected:
image
- It appears twice when "Manually fill in" is selected:
image

src/components/builder/values/values-config.tsx
```diff
@@ -1,11 +1,26 @@
...
 import {useLayoutEffect} from 'react';
 
+import {hasOwnProperty} from '@/types';
+
 import ItemsExpression from './items-expression';
 import {SchemaWithDataSrc} from './types';
 import ValuesSrc from './values-src';
 import ValuesTable, {ValuesTableProps} from './values-table';
 
+function isNestedKeySet(obj: {}, path: string): boolean {
```
Viicos, line 11:
```suggestion
/**
 * Check if the dotted `path` exists on `obj`.
 * @example
 * ```
 * isNestedKeySet({my: {path: true}}, 'my.path') // true
 * ```
 */
function isNestedKeySet(obj: {}, path: string): boolean {
```

src/registry/select/edit.tsx
```diff
@@ -0,0 +1,206 @@
...
+  Validate.useManageValidatorsTranslations<SelectComponentSchema>(['required']);
+
+  const isManualOptions = checkIsManualOptions(values);
+  const options = isManualOptions ? values.data?.values || [] : [];
+
+  // Ensure that form state is reset if the values source changes.
+  useLayoutEffect(() => {
+    if (dataSrc !== 'variable' || isEqual(defaultValue, {})) return;
+    setFieldValue('defaultValue', '');
+  }, [dataSrc]);
```
Viicos, line 66: Should we watch for `defaultValue` as well?

Viicos (approved):

sergei-maertens:
> I am experiencing a weird behavior with datasource:
> 
> * a `*` appears in the select when "From variable" is selected:
>   image
> * It appears twice when "Manually fill in" is selected:
>   image

It's a CSS issue with bootstrap/formio that applies a dumb z-index to the `required` indicator of the field below (either values table or itemsexpression). I don't know if this issue is also present in our own admin since we've done a lot of CSS fixes like that there already :thinking:

src/registry/select/edit.tsx
sergei-maertens: I don't think so, because the reset behaviour should only kick in when the `dataSrc` value changes.

If the `dataSrc` is set to `manual` and the user changes the default value in the dropdown, this effect would cause it to be reset to no default value which entirely beats the point of having a default value field :P

There is a bug here however - the comparison should be against `''` instead of `{}` in the guard clause.

The idea is that whenever the user:

* switch from variable -> manual: nothing to do, since it's not possible to set a default value for 'variable' dataSrc
* switch from manual -> variable: if a default value was set, clear it, since we don't know which values will exist (they come from `itemsExpression`). if no default value was set -> nothing to do/reset

sergei-maertens: This made me realize that the `multiple` behaviour was also not incorporated - which is not relevant for selectboxes or radio, so I've included that in my latest commit.

Viicos: But what about ESLint in this case? As you said it requires every variable used to be watched?

sergei-maertens: yeah, this would be one case where you silence the ESLint warning - complying with the ESLint rule would lead to incorrect behaviour in this case.

It's not a perfect linter rule, unfortunately. The goal of React was also to figure out the dependencies at compile time, but they can't do so reliably (yet) because of usages like this.

We have another case like this in the SDK, where progress indicator on mobile should collapse on navigate: https://github.com/open-formulieren/open-forms-sdk/blob/8117bacd777d785b197703cc9948bce9c9e004fc/src/components/ProgressIndicator/index.js#L30

Letting the hook run when `expanded` changes there basically prevents you from ever opening the progress indicator. Sometimes a certain state should deliberately be excluded from the hook triggers.

PR #68, opened 2023-11-24
Add new `address` component 

### LINKED ISSUE #65: Implement new `address` component
Part of https://github.com/open-formulieren/open-forms/issues/3607

Viicos:
Fixes #65 
Needs https://github.com/open-formulieren/types/pull/30/

This might need additional user configuration:
- configurable labels for each input?

### DIFF
--- a/src/registry/address/edit-validation.ts
+++ b/src/registry/address/edit-validation.ts
@@ -2,13 +2,20 @@ import {IntlShape} from 'react-intl';
 import {z} from 'zod';
 
 import {buildCommonSchema} from '@/registry/validation';
+
 import {POSTCODE_REGEX} from '../postcode/constants';
 
 // Constraints taken from the BRK API (apart from postcode which comes from our postcode component)
 const addressSchema = z.object({
-  postcode: z.string().regex(new RegExp(POSTCODE_REGEX)),
-  houseNumber: z.string().regex(/^\d{1,5}$/),
-  houseLetter: z.string().regex(/^[a-zA-Z]$/),
+  postcode: z.string().regex(new RegExp(POSTCODE_REGEX)).optional(),
+  houseNumber: z
+    .string()
+    .regex(/^\d{1,5}$/)
+    .optional(),
+  houseLetter: z
+    .string()
+    .regex(/^[a-zA-Z]$/)
+    .optional(),
   houseNumberAddition: z
     .string()
     .regex(/^([a-z,A-Z,0-9]){1,4}$/)


src/components/ComponentConfiguration.stories.tsx
```diff
@@ -1691,3 +1691,59 @@ export const BSN: Story = {
     expect(args.onSubmit).toHaveBeenCalled();
   },
 };
+
+export const Address: Story = {
+  render: Template,
+  name: 'type: address',
+
+  args: {
```
sergei-maertens, line 1699: `builderInfo` arg is missing, so now the title displays "text field component"

src/components/ComponentConfiguration.stories.tsx
```diff
@@ -1691,3 +1691,59 @@ export const BSN: Story = {
...
+
+    // ensure that changing fields in the edit form properly update the preview
+    const preview = within(canvas.getByTestId('componentPreview'));
+
+    await userEvent.clear(canvas.getByLabelText('Label'));
+    await userEvent.type(canvas.getByLabelText('Label'), 'Updated preview label');
+    expect(await preview.findByText('Updated preview label'));
+
+    const previewInput = preview.getByText('Updated preview label');
+    await expect(previewInput).toHaveDisplayValue('');
```
sergei-maertens, line 1732:
```suggestion
    await expect(previewInput).toBeVisible();
```

it's not an input field

src/components/formio/fieldset.tsx
sergei-maertens, line 1: please add separate stories for this newly added component

src/components/formio/fieldset.tsx
```diff
@@ -0,0 +1,64 @@
...
+  type,
+  field = '',
+  required = false,
+  label,
+  tooltip = '',
+  children,
+}) => {
+  const {errors} = useValidationErrors(field);
+  const className = clsx('form-group', 'has-feedback', 'formio-component', PAD_ERROR_ICON, {
+    [`formio-component-${type}`]: type,
```
sergei-maertens, line 41: do we need this class name? there's definitely no built in formio styles targetting this since it's a custom component

src/components/formio/fieldset.tsx
```diff
@@ -0,0 +1,64 @@
...
+    'has-error': field && errors.length > 0,
+    required: required,
+  });
+  // const labelClassName = clsx('col-form-label', {'field-required': required});
```
sergei-maertens, line 45:
```suggestion
```

src/registry/address/edit-validation.ts
```diff
@@ -0,0 +1,21 @@
...
+  postcode: z.string().regex(/^[1-9][0-9]{3} ?(?!sa|sd|ss|SA|SD|SS)[a-zA-Z]{2}$/),
+  houseNumber: z.string().regex(/^\d{1,5}$/),
+  houseLetter: z.string().regex(/^[a-zA-Z]$/),
+  houseNumberAddition: z
+    .string()
+    .regex(/^([a-z,A-Z,0-9]){1,4}$/)
+    .optional(),
+});
+
+const defaultValueSchema = z.object({defaultValue: addressSchema});
```
sergei-maertens, line 17: this is defined, but not exposed anywhere, so you can now have validation errors that are not rendered anywhere. You can leave it as is, since no inputs also means you can't enter invalid values.

sergei-maertens (changes requested):

src/registry/address/edit-validation.ts
Viicos: Yeah even having the `addressSchema` doesn't seem that relevant as we don't allow default values.

src/registry/address/edit-validation.ts
```diff
@@ -0,0 +1,21 @@
+import {IntlShape} from 'react-intl';
+import {z} from 'zod';
+
+import {buildCommonSchema} from '@/registry/validation';
+
+// Constraints taken from the BRK API (apart from postcode which comes from our postcode component)
+const addressSchema = z.object({
+  postcode: z.string().regex(/^[1-9][0-9]{3} ?(?!sa|sd|ss|SA|SD|SS)[a-zA-Z]{2}$/),
```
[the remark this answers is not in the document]
Viicos: It is defined as a constant in the `postcode` component, so I could import from there (though postcode is deprecated, so not sure about that?)

src/components/formio/fieldset.stories.tsx
```diff
@@ -0,0 +1,69 @@
...
+  title: 'Formio/Containers/Fieldset',
+  component: Fieldset,
+  decorators: [withFormik],
+  parameters: {
+    modal: {noModal: true},
+    // https://github.com/bbbtech/storybook-formik/issues/51#issuecomment-1136668271
+    docs: {
+      inlineStories: false,
+      iframeHeight: 150,
+    },
```
sergei-maertens, line 19: no longer relevant since this dependency was dropped :)

src/components/formio/fieldset.stories.tsx
```diff
@@ -0,0 +1,69 @@
...
+  },
+} as Meta<typeof Fieldset>;
+
+type Story = StoryObj<typeof Fieldset>;
+
+const Template: StoryFn<typeof Fieldset> = args => (
+  <Fieldset {...args}>
+    <div>{args.children || '<any children />'}</div>
+  </Fieldset>
+);
```
sergei-maertens, line 32: I think you can drop this entirely and just set the `children` arg default on the `default` export to `<any children />` string

src/components/formio/fieldset.stories.tsx
```diff
@@ -0,0 +1,69 @@
...
+  children: <input type="text" />,
+  field: 'someField',
+};
+WithErrors.argTypes = {
+  children: {table: {disable: true}},
+};
+WithErrors.play = async ({canvasElement}) => {
+  const canvas = within(canvasElement);
+  await expect(canvas.queryByText('Some error')).toBeInTheDocument();
+};
```
sergei-maertens, line 69: Can you please use CSF 3 intead of CSF2?

src/registry/address/edit-validation.ts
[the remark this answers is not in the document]
sergei-maertens: sure, import it from there :)

Viicos:
Regarding the comments on the fieldset stories, it is mostly copy pasted from the component stories, which would probably need to be also updated

src/registry/address/edit-validation.ts
```diff
@@ -0,0 +1,29 @@
...
+    .regex(/^\d{1,5}$/)
+    .optional(),
+  houseLetter: z
+    .string()
+    .regex(/^[a-zA-Z]$/)
+    .optional(),
+  houseNumberAddition: z
+    .string()
+    .regex(/^([a-z,A-Z,0-9]){1,4}$/)
+    .optional(),
```
Viicos, line 22: Is something like `z.union([z.string().(...), z.literal('')])` more accurate?

sergei-maertens: Not really, there's a nasty implementation detail that's relevant for the validation: when the input has no value (`''` in this case), formik will change that to `undefined` before passing it to validation. While this is counter-intuitive, the upstream issue/reason for this makes sense and it's something we have to live with. So the `z.literal('')` case will never actually happen, instead this ends up in the `.optional()` flow.

So what you currently have is the best reflection of the form inputs + validation. From a pure data modelling perspective, it could be better, but that's not what we are doing here.

sergei-maertens (approved):

PR #69, opened 2023-11-28
Add type checking to items expression

CharString:
Closes https://github.com/open-formulieren/open-forms/issues/3596
Closes https://github.com/open-formulieren/open-forms/issues/3597


src/context.ts
```diff
@@ -47,6 +48,43 @@ export interface DocumentTypeOption {
   };
 }
 
+export interface ServiceFetchConfiguration {
```
sergei-maertens, line 51: hrm, I'm not sure this belongs in the form-builder repository - does something break if you ignore `serviceFetchConfiguration` in the `VariableDefinition` type?

Perhaps it makes more sense to pass `getInfernoLogicContext` via the builder context rather than providing the inputs and creating this context? That way we can build that context in our backend JS and pass it to the builder, and the open-forms specific stuff is more decoupled from the formio domain for the builder? Ultimately you sort of initialize the builder with "here, this is the infernologic type checking context, use that for your type checks and please validate the things in the builder" without the builder itself needing to understand what `infer` needs (other than the function signature).

src/components/builder/values/items-expression.tsx
```diff
@@ -1,12 +1,96 @@
...
+import {useContext} from 'react';
 import {FormattedMessage} from 'react-intl';
 
 import JSONEdit from '@/components/JSONEdit';
 import {Component, Description} from '@/components/formio';
+import {BuilderContext, VariableDefinition} from '@/context';
 
 const NAME = 'openForms.itemsExpression';
 
+const dataTypeForComponent = (component: AnyComponentSchema): JSONValue => {
```
sergei-maertens, line 13: would it make sense to add this to the `infernologic` library itself? In a `contrib/formio` package or something like that? That would also make it more easily shared between the builder & backend.

src/components/JSONEdit.tsx
```diff
@@ -1,25 +1,28 @@
-import {JSONObject} from '@open-formulieren/types/lib/types';
+import {type JSONObject, type JSONValue} from '@open-formulieren/types/lib/types';
```
sergei-maertens, line 1: not saying this is wrong, but why is this `type` addition needed?

src/components/JSONEdit.tsx
```diff
@@ -39,26 +42,39 @@ const JSONEdit: React.FC<JSONEditProps & TextareaHTMLAttributes<HTMLTextAreaElem
     let updatedData: any;
     try {
       updatedData = JSON.parse(rawValue);
-      setJSONValid(true);
-    } catch {
-      setJSONValid(false);
+      setJsonLogicError('');
+    } catch (error) {
+      setJsonLogicError(error.toString());
```
sergei-maertens, line 47: is this properly localized?

src/components/JSONEdit.tsx
```diff
@@ -39,26 +42,39 @@ const JSONEdit: React.FC<JSONEditProps & TextareaHTMLAttributes<HTMLTextAreaElem
...
       return;
     }
-
-    updateValue(updatedData);
+    const error = validateLogic(updatedData);
+    if (error) {
+      setJsonLogicError(error);
+      return;
+    }
+    updateValue(updatedData); // valid JSON
```
sergei-maertens:
```suggestion
    updateValue(updatedData); // valid JSON & valid logic expression
```
because of `validateLogic` returns an error message, we don't run `updateValue` so the inverse must be true :)

src/components/JSONEdit.tsx
```diff
@@ -39,26 +42,39 @@ const JSONEdit: React.FC<JSONEditProps & TextareaHTMLAttributes<HTMLTextAreaElem
...
   };
+
+  const errorMessageId = JsonLogicError ? uniqueId() : '';
```
sergei-maertens, line 58:
```suggestion
    // TODO React 18: use React.useId()
    const errorMessageId = JsonLogicError ? uniqueId() : '';
```

src/components/builder/values/items-expression.tsx
```diff
@@ -32,7 +36,13 @@ export const ItemsExpression: React.FC = () => {
       }
     >
       <div>
-        <JSONEdit name={NAME} data={value} rows={3} id={htmlId} />
+        <JSONEdit
+          name={NAME}
+          data={value}
+          rows={3}
+          id={htmlId}
+          validateLogic={logic => validateLogic(logic, [['', '']])}
```
sergei-maertens, line 44: IMO this validation should not be passed at the component level, but rather at the zod schema level. Alternatively, you could pass it in at the Formik field level: https://formik.org/docs/api/field#validate

But the end result is that validation errors from field-level validation and form-level validation is the same - they block submitting the form because it is not valid, and the validation errors are normalized inside the `errors[fieldName]` data structure (where `errors`, `values` and `touched` all share the same structure w/r to expanding the `.` etc.).

src/utils/jsonlogic.ts
```diff
@@ -0,0 +1,169 @@
...
+ * JsonLogic type checking utility functions
+ */
+import {infer} from '@open-formulieren/infernologic';
+import {type JSONObject, type JSONValue} from '@open-formulieren/types/lib/types';
+
+import {type AnyComponentSchema} from '@/types';
+
+/**
+ * @param {object | array | number | string} logic - JsonLogic expression
+ * @param {object | array | number | string} expected - example value from expected result type e.g. [["label", "value"]]
```
sergei-maertens, line 11: I've never understood how the JSDoc type definitions stack up against the actual type definitions in the code itself. To me it seems a risk of going out of sync.

The types here don't account for `null` which is a valid JSON type and I can't predict how things behave if I pass `null` (since the TS compiler will likely allow it).

Additionally, the `object` will probably allow me to pass `function` references :)

src/utils/jsonlogic.ts
```diff
@@ -0,0 +1,169 @@
+/**
+ * JsonLogic type checking utility functions
+ */
```
sergei-maertens, line 3: maybe a `@todo` is in order to figure out a better place for these type definitions? I regrettably must admit that using a monorepo for all our TS would probably make sense to put code in logical places while still being able to import/package it to well-scoped packages. Having this OF backend specific stuff in something that pretends to talk "only" formio concepts is blurring the lines.

src/utils/jsonlogic.ts
```diff
@@ -0,0 +1,169 @@
...
+  | 'boolean';
+
+interface ServiceFetchConfiguration {
+  id: number;
+  name: string;
+  service: string;
+  path: string;
+  method: 'GET' | 'POST';
+  headers: Record<string, string>;
+  query_params: Record<string, string>;
```
sergei-maertens, line 34:
```suggestion
  query_params: Record<string, string[]>;
```
I think, because we have this array value construction: https://open-formulieren.github.io/open-forms/?path=/docs/admin-custom-mappingarrayinput--docs#mapping-to-arrays-using-valuearrayinput-argument

src/utils/jsonlogic.ts
```diff
@@ -0,0 +1,169 @@
...
+
+/**
+ * Return a function that can type check whether
+ */
+export const createTypeCheck = ({
+  formVariables = [],
+  staticVariables = [],
+  components = [],
+}: FormContext = {}): JsonLogicTypeChecker => {
+  const variableContext = staticVariables.concat(formVariables);
```
sergei-maertens, line 77: aha, found it :)

src/utils/jsonlogic.ts
```diff
@@ -0,0 +1,169 @@
...
+        datetime: '',
+        time: '',
+        boolean: true,
+      }[dataType];
+
+const dataTypeForComponent = (
+  component:
+    | AnyComponentSchema
+    | {
+        type: 'map' | 'editgrid' | 'password' | 'signature'; // Not yet implemented in this builder
```
sergei-maertens, line 116: can you add some `// TODO` or `// XXX` or `// FIXME` markers? :)

and more are coming: ~~`address`~~, `cosign` and `coSign`, I think that's all of them.

src/utils/jsonlogic.ts
```diff
@@ -0,0 +1,169 @@
...
+        multiple?: boolean;
+        defaultValue: JSONValue;
+      }
+): JSONValue => {
+  // For now return example values as accepted by InferNoLogic
+  // But example values cannot distinguish arrays from tuples!
```
sergei-maertens, line 122: not even when using `[...] as const`?

src/utils/jsonlogic.ts
```diff
@@ -0,0 +1,169 @@
...
+      houseNumber: '',
+      houseLetter: '',
+      houseNumberAddition: '',
+    },
+    currency: 1,
+    number: 1,
+    checkbox: true,
+    //@ts-ignore selectboxes always have component.defaultValue (this does work when rewritten as a lengthy switch/case)
+    selectboxes: component.defaultValue,
+    npFamilyMembers: {}, // TODO record type
```
sergei-maertens, line 135: this ultimately becomes the same as selectboxes BUT it can also become a `type: 'content'` component if there's an error :grimacing:

src/utils/jsonlogic.ts
```diff
@@ -0,0 +1,169 @@
...
+    phoneNumber: 'string',
+    postcode: 'string',
+    radio: 'string',
+    select: 'string',
+    signature: 'string', // data-url
+    textarea: 'string',
+    textfield: 'string',
+    time: 'time', // TODO string for now
+  }[component.type];
+  return 'multiple' in component && component.multiple ? [value] : value;
```
sergei-maertens, line 168: file multiple is still just an array of objects, I think this produces `[[{...}]]` now?

Also a more correct check is `hasOwnProperty(component, 'multiple')` which also adds the type marker that `component.multiple` actually exists. It prevents some accidents in the prototype chain where `multiple` could be defined as a method for example, which has a different meaning.

src/components/ComponentConfiguration.stories.tsx
```diff
@@ -69,6 +72,7 @@ export default {
       schema: {placeholder: ''},
       weight: 0,
     },
+    onSubmit: jest.fn(),
```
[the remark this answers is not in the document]
CharString: It was a red herring. Just silences a warning, now.

src/components/JSONEdit.tsx
CharString: It's not needed here, I think. That's it. Always explicitly importing types as type declarations makes me not have to think ever, peek into the module I'm importing from and reasoning if it will ever turn into an [edge case](https://devblogs.microsoft.com/typescript/announcing-typescript-3-8-beta/#type-only-imports-exports).

I guess it's also easier on the compiler, but it's mostly easier on my own brain cycles.

src/components/JSONEdit.tsx
CharString: No. But neither are the JSON.parse() errors. Discussed it with Joeri. L10n could be something for the future.

src/utils/jsonlogic.ts
CharString: Yeah. using `hasOwnProperty` doesn't type check;

.storybook/decorators.tsx
```diff
@@ -216,13 +218,16 @@ export const BuilderContextDecorator = (Story: StoryFn, context: StoryContext) =
...
+  const staticVariables = static_variables as VariableDefinition[]; // source is inferred as string not  as ""
   return (
     <BuilderContext.Provider
       value={{
         uniquifyKey: key => key,
         supportedLanguageCodes: supportedLanguageCodes,
         componentTranslationsRef: {current: translationsStore},
-        getFormComponents: () => context?.args?.componentTree || defaultComponentTree,
+        getFormComponents: () => components,
+        validateLogic: createTypeCheck({components, staticVariables}),
```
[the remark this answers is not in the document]
Viicos: Seems to be, although this is for storybook, so we only load the static variables.

src/components/JSONEdit.tsx
```diff
@@ -46,6 +46,7 @@ const JSONEdit: React.FC<JSONEditProps & TextareaHTMLAttributes<HTMLTextAreaElem
     }
 
     updateValue(updatedData);
+    name && validateField(name);
```
Viicos, line 49: @sergei-maertens we are hitting an issue here and don't know how to get things working:

`name && validateField(name)` was added because the `updateValue` call above did not trigger the `validate` callback [here](https://github.com/open-formulieren/formio-builder/pull/69/files#diff-916b5a36551dfa9c19ff511ea1397be46e02d9e474a85121f968c4bd7c132581R42) (even though `validateOnChange` is set to `true`).

The issue is data gets out of sync, as here `updatedData` corresponds to the latest version of the user input data, whereas adding a `console.log` in the parent `validate` callback shows an outdated value.

I've tried adding a call to the base Formik `onChange`:

```ts
const {onChange: formikOnChange} = getFieldProps(name);
// Then calling `formikOnChange` in our custom `onChange`
```

But nothing works out

src/utils/jsonlogic.ts
Viicos: What other components would make use of this util module?

src/utils/jsonlogic.ts
Viicos: Not sure what Chris and you meant here

src/components/JSONEdit.tsx
sergei-maertens: I still don't think that wiring up this validate behaviour directly in the component is the right approach. Especially because this JSONEdit component is also used to edit the entire form component, but it could also point at a different subtree that has nothing to do with JSONLogic.

IMO hooking up the validator belongs in the zod validation schema, where you can make use of `.refine` or `.superRefine` to invoke this validator:

```tsx
const schema = z.object({
  itemsExpression: z.record(z.unknown()).superRefine(
    ...  // in the callback here you'd call the `infer` validator/checker
  ),
});
```

src/utils/jsonlogic.ts
sergei-maertens: we want to use TS for our admin-builder code too, which is currently webpack/JS. So these type definitions will become relevant for the backend project. The whole service fetch configuration modal/form components would benefit from these types.

src/utils/jsonlogic.ts
sergei-maertens: I think the issue is that some expressions like `['foo, 'bar']` get inferred as `string[]` where-as they should be a two-tuple. If you define a fixed-length array of types, then it becomes a tuple in TS: https://www.typescriptlang.org/docs/handbook/basic-types.html#tuple

I was wondering if using the `value as const` construct would make TS infer it as a tuple rather than an `Array`

src/utils/jsonlogic.ts
Viicos: Ok I see, admin-builder == the current js in the django backend I suppose

src/components/JSONEdit.tsx
Viicos: This is cleaner indeed. Unfortunately I seem to be facing https://github.com/robertLichtnow/zod-formik-adapter/issues/15, will have to look closer.

src/utils/jsonlogic.ts
CharString: The problem is not TS, but open-formulieren/InferNoLogic/issues/5

CharString: At some point the returntype of this function will have to become some form of type annotation. Not just to distinguish `array` from `tuple`, but also to distinguish `string` from `date`, `datetime` etc.

src/components/JSONEdit.tsx
sergei-maertens: That seems weird since the (super)refine is used in the columns component validation.

Viicos: I see it being called with some logs, but the error doesn't seem to be displayed.

sergei-maertens:
Sorry, postponing this :(

Issue #81, opened 2023-12-08
Set up Jest/testing library for tests


sergei-maertens:
As an alternative to running (integration/unit) tests in Storybook, which is too flaky.

sergei-maertens:
Completed via #82

PR #82, opened 2023-12-08
[#81] Initial test setup

Viicos:


src/test-utils.tsx
sergei-maertens, line 1: as it stands, this file would be compiled and shipped in the `dist/` or `lib/` folder, right?

tsconfig.json
```diff
@@ -30,8 +30,9 @@
         "overrideIdFn": "[sha512:contenthash:base64:6]",
         "ast": true
       }
-    ]
+    ],
+    "types": ["node", "jest", "@testing-library/jest-dom"],
   },
   "include": ["src"],
-  "exclude": ["node_modules", "lib"]
+  "exclude": ["node_modules", "lib"],
```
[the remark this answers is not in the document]
Viicos: Typescript supports comments and commas, but yes better stick with the standard if this is ever parsed by an external tool

src/components/ComponentEditForm.tsx
```diff
@@ -27,18 +27,18 @@ export interface ComponentEditFormProps {
   // don't know (yet), so we need to handle FallbackSchema.
   component: AnyComponentSchema | FallbackSchema;
   builderInfo: BuilderInfo;
-  onCancel: (e: React.MouseEvent<HTMLButtonElement>) => void;
-  onRemove: (e: React.MouseEvent<HTMLButtonElement>) => void;
-  onSubmit: (component: AnyComponentSchema | FallbackSchema) => void;
+  onCancel?: (e: React.MouseEvent<HTMLButtonElement>) => void;
+  onRemove?: (e: React.MouseEvent<HTMLButtonElement>) => void;
+  onSubmit?: (component: AnyComponentSchema | FallbackSchema) => void;
```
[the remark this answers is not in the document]
Viicos: There's a lot of SB components that do not provide them, but ts did not complain as `args` are not typechecked:

```ts
export default {
  component: ComponentEditForm,
  args: ...  // `args` not type checked against `component`
} as Meta<typeof ComponentEditForm>;
```

But when writing the Jest test, it does typecheck and noticed these three props where not provided.

src/test-utils.tsx
```diff
@@ -0,0 +1,304 @@
...
+          </Formik>
+        )}
+      </IntlProvider>
+    );
+  }
+  return rtlRender(ui, {wrapper: Wrapper, ...renderOptions});
+};
+
+// re-export everything
+export * from '@testing-library/react';
```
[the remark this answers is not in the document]
Viicos: This is the recommended way of Jest: https://testing-library.com/docs/react-testing-library/setup/#custom-render

So you can import everything from `test-utils`. Don't know if this is what we want though.

src/registry/select/select-validation.spec.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+  contextRender(<ComponentEditForm {...args} />);
+
+  test('Option values and labels are required fields', async () => {
+    const labelInput = screen.getByLabelText('Option label');
+    await userEvent.type(labelInput, 'Foo');
+    await userEvent.clear(labelInput);
+    await userEvent.keyboard('[Tab]');
+    await expect(await screen.findByText('The option label is a required field.')).toBeVisible();
+    await expect(await screen.findByText('The option value is a required field.')).toBeVisible();
```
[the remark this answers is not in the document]
Viicos: You mean moving `contextRender` to the `test` call?

package.json
```diff
@@ -19,6 +19,7 @@
     "build:cjs": "ttsc -p tsconfig.prod.json --module commonjs --outDir lib/cjs && tsc-alias -p tsconfig.prod.json --outDir lib/cjs",
     "build": "npm-run-all build:*",
     "test": "test-storybook",
+    "test-jest": "jest",
```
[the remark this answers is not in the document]
Viicos: Does this mean we keep all SB tests and duplicate them as Jest tests?

src/components/ComponentEditForm.tsx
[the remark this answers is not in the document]
sergei-maertens: Storybook automatically injects spies for the `on*` pattern, that's why it type checks. SB 8 will make this explicit, using `import {fn} from '@storybook/test'`, but we're not there yet.

Storybook also applies `Partial<Props>` afaik for the stories, so keeping it required and passing explicit `jest.fn()` as props in jest tests seems like the better approach to me.

package.json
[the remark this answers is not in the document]
sergei-maertens: No, some of these (interaction) tests really belong in storybook where as other (unit, integration) are better suited for jest. The line is blurry, but the advantage of having some tests in storybook  is that the intended interactions are very clear *and* they're visually obvious what is going wrong. With jest tests, debugging tests tends to be quite painful since you don't know what's actually on the browser screen.

src/registry/select/select-validation.spec.tsx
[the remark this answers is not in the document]
sergei-maertens: yep

src/test-utils.tsx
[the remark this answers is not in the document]
sergei-maertens: I see - please include this link/reference as a comment in the code.

tsconfig.json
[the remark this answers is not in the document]
sergei-maertens: Yeah I've got some plugins crashing in my editor on config files with trailing comma's because they use the Python API which does not allow trailing comma's.

I'm okay with trailing comma's/comments if the appropriate extension is used (like https://json5.org/), but in this case best to just stick to plain JSON.

.github/workflows/ci.yml
```diff
@@ -86,6 +86,26 @@ jobs:
...
+      - name: Install dependencies
+        run: npm install
+
+      - name: Run tests
+        run: |
+          npm run test:jest
+        env:
+          CI: 'true'
+
   tests:
```
sergei-maertens, line 109: I'd rename this:

```suggestion
  tests-storybook:
```

.github/workflows/ci.yml
```diff
@@ -86,6 +86,26 @@ jobs:
       - name: Run prettier linter
         run: npm run checkformat
 
+  tests-jest:
```
sergei-maertens, line 89: this key needs to be added to the `needs` array in the `deploy` job :) don't want to deploy if there are any test failures, jest or storybook

.github/workflows/ci.yml
```diff
@@ -86,6 +86,26 @@ jobs:
...
+          cache: npm
+
+      - name: Install dependencies
+        run: npm install
+
+      - name: Run tests
+        run: |
+          npm run test:jest
+        env:
+          CI: 'true'
```
sergei-maertens, line 107: I thought this is detected automatically, is it not the case?

babel.config.js
sergei-maertens, line 1: why does this work and `.babelrc` doesn't?

src/registry/select/select-validation.spec.tsx
```diff
@@ -0,0 +1,46 @@
...
+  };
+  const builderInfo = {
+    title: 'Select',
+    icon: 'th-list',
+    group: 'basic',
+    weight: 70,
+    schema: {},
+  };
+
+  test('Option values and labels are required fields', async () => {
```
sergei-maertens, line 27: generally we write the test descriptions so that `describe + test/it` creates a logical English sentence - any way you can improve on that?

src/registry/select/select-validation.spec.tsx
```diff
@@ -0,0 +1,46 @@
...
+    values: [{value: '', label: ''}],
+    defaultValue: '',
+  };
+  const builderInfo = {
+    title: 'Select',
+    icon: 'th-list',
+    group: 'basic',
+    weight: 70,
+    schema: {},
+  };
```
sergei-maertens, line 25: I can't wait for this to be centralized :)

You can add some type annotations though: `const component = {...} satisfies SelectComponentSchema;`

src/test-utils.tsx
```diff
@@ -0,0 +1,304 @@
...
+
+import {BuilderContext} from '@/context';
+import type {ComponentTranslationsRef, DocumentTypeOption, SelectOption} from '@/context';
+import {AnyComponentSchema} from '@/types';
+
+import {PrefillAttributeOption, PrefillPluginOption} from '../src/components/builder/prefill';
+import {RegistrationAttributeOption} from '../src/components/builder/registration/registration-attribute';
+import {ValidatorOption} from '../src/components/builder/validate/validator-select';
+
+const DEFAULT_COMPONENT_TREE: AnyComponentSchema[] = [
```
sergei-maertens, line 17: I think it'd be wise to define these constants in place shared between SB & jest? We're still adding some context callables that rely on stuff being put in and having to update two files with the same dummy data is annoying.

package.json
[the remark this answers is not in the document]
Viicos: Ok makes sense

babel.config.js
Viicos: Because the clever people at Babel decided that depending on the configuration file used, it will behave differently: https://babeljs.io/docs/configuration

In my case, here is the relevant info:
- https://github.com/jestjs/jest/issues/6229#issuecomment-392205646
- https://github.com/jestjs/jest/issues/6053#issuecomment-383632515

src/test-utils.tsx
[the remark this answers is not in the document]
Viicos: Yes this is currently the case, I'm exporting two other render functions `contextRender` and `formikRender`, but the base `render` is still being export by this line.

.github/workflows/ci.yml
Viicos: I don't remember where I took this from

src/registry/select/select-validation.spec.tsx
Viicos: Hum, `Object literal may only specify known properties, and 'values' does not exist in type 'SelectComponentSchema'`. Apparently `values` should be nested under `data` according to the type definition. But I've just copied the structure from the SB test, so it might be invalid there as well?

sergei-maertens: It likely is - the component edit form itself applies default values which may shadow this issue then. Please do check it.

Viicos:
Applied remaining review. SB tests for the cosign component fails (even locally, so not related to flakyness).

Once this PR is approved, I think it would be nice to merge it and move tests in another PR.

sergei-maertens (approved):
I'll take a look at the test failures in a bit

src/tests/test-utils.tsx
```diff
@@ -0,0 +1,145 @@
...
+  defaultConfidentialityLevels: SelectOption[];
+  registrationAttributesDelay: number;
+}
+
+const contextRender = (
+  ui: React.ReactElement,
+  enableContext: boolean = true,
+  locale: string = 'en',
+  builderOptions: Partial<BuilderOptions> = {},
+  renderOptions: RenderOptions = {}
```
sergei-maertens, line 48: discussed in the office, the signature should be updated to `contextRender(ui: React.ReactElement, opts: {...})`

sergei-maertens (approved):

PR #85, opened 2023-12-11
Implement cosign components

### LINKED ISSUE #84: Implement cosign component edit forms

sergei-maertens:
Closes #84

- [x] Implement cosign v1
- [x] Implement cosign v2

### DIFF
--- a/src/registry/cosignV1/edit.tsx
+++ b/src/registry/cosignV1/edit.tsx
@@ -1,0 +1,130 @@
+import {CosignV1ComponentSchema} from '@open-formulieren/types';
+import {useFormikContext} from 'formik';
+import {useContext} from 'react';
+import {FormattedMessage, useIntl} from 'react-intl';
+import useAsync from 'react-use/esm/useAsync';
+
+import {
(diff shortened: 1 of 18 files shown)


src/registry/cosignV1/edit.tsx
```diff
@@ -0,0 +1,129 @@
...
+        <FormattedMessage
+          description="Label for 'authPlugin' builder field"
+          defaultMessage="Authentication method"
+        />
+      }
+      description={
+        <FormattedMessage
+          description="Description for 'authPlugin' builder field"
+          defaultMessage={`Which authentication method the co-signer must use. Note that
+          this must be an authentication method available on the form.`}
```
Viicos, line 119: Why not use a <kbd>?</kbd> instead? Is it to have the note displayed to the user by default?

sergei-maertens: Correct. It contains an important constraint that should always be visible, as it affects what is valid and what is not, whereas a tooltip is usually informational.

Viicos (approved):

PR #90, opened 2023-12-14
Implement map component

### LINKED ISSUE #89: Implement map component in builder

sergei-maertens:
Closes #89 
Builds on top of #88


.storybook/preview.ts
```diff
@@ -4,10 +4,36 @@ import '@fortawesome/fontawesome-free/css/v4-shims.css';
...
+  // @ts-ignore
+  delete L.Icon.Default.prototype._getIconUrl;
+  // the call to `require` ensures that the static assets are bundled along (or, for small
+  // images, inlined as base64)
+  L.Icon.Default.mergeOptions({
+    iconRetinaUrl: iconRetinaUrl,
+    iconUrl: iconUrl,
+    shadowUrl: shadowUrl,
+  });
+};
```
Viicos, line 33: I remember having to do the same a couple years ago on a Vue project with leaflet as a dep., funny to see this is still something

Viicos (approved):

.storybook/preview.ts
sergei-maertens: Yeah leaflet authors themselves do not want to fix anything in their code base, and I get it, since it's specific to how bundlers treat this stuff.

I'm not even sure things work here or if this should be shipped in the builder, since we're not displaying a pin anyway. I'll take another look.

PR #93, opened 2023-12-15
Upgrade to React 18

### LINKED ISSUE #92: Upgrade to React 18
Now the backend has upgraded to React 18, this project should be able too.

sergei-maertens:
Closes #92

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,7 @@
     "@storybook/test-runner": "^0.16.0",
     "@storybook/testing-library": "^0.2.2",
     "@testing-library/jest-dom": "^6.1.5",
-    "@testing-library/react": "^12.1.5",
+    "@testing-library/react": "^14.1.2",
     "@trivago/prettier-plugin-sort-imports": "^4.1.1",
     "@types/leaflet": "^1.9.8",
     "@types/lodash.camelcase": "^4.3.7",
@@ -85,7 +85,7 @@
     "@types/lodash.merge": "^4.6.7",
     "@types/lodash.uniqueid": "^4.0.7",
     "@types/proj4leaflet": "^1.0.10",
-    "@types/react": "^18.0.33",
+    "@types/react": "^18.2.45",
     "@types/react-signature-canvas": "^1.0.5",
     "@types/sanitize-html": "^2.9.0",
     "@types/signature_pad": "^2.3.6",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -85,7 +85,7 @@
     "@types/lodash.merge": "^4.6.7",
     "@types/lodash.uniqueid": "^4.0.7",
     "@types/proj4leaflet": "^1.0.10",
-    "@types/react": "^18.0.33",
+    "@types/react": "^18.2.45",
```
Viicos, line 88: I always find it confusing when types packages' version seems to be aligned with the actual package. Is `@types/react@18.2.x` compatible with `react@18.2.x`? I guess so

src/components/builder/prefill/prefill.stories.tsx
```diff
@@ -60,19 +60,25 @@ export const PluginBlankNoAttributes: Story = {
...
+
+    await waitFor(
+      async () => {
+        await userEvent.keyboard('[Esc]');
+        pluginInput.focus();
+        await userEvent.keyboard('[ArrowDown]');
+        expect(canvas.queryByText('Plugin 1')).toBeInTheDocument();
+        expect(canvas.queryByText('Plugin 2')).toBeInTheDocument();
+      },
+      {timeout: 1500, container: canvasElement}
```
Viicos, line 73: I hope we will be able to get rid of this arbitrary timeout when moved to jest

src/components/builder/registration/registration-attribute.spec.tsx
```diff
@@ -38,8 +38,10 @@ test('Available registration attributes are retrieved via context', async () =>
 
   // open the dropdown
   const input = await screen.findByLabelText('Registration attribute');
-  input.focus();
-  await user.keyboard('[ArrowDown]');
+  await act(async () => {
+    input.focus();
+    await user.keyboard('[ArrowDown]');
+  });
```
Viicos, line 44: I'm confused: is this `act` call required to suppress the warning? If so, `act` is being called to times, as testing-library apparently handle it already?

Viicos (approved):

package.json
sergei-maertens: I think and hope so :grimacing:

src/components/builder/prefill/prefill.stories.tsx
sergei-maertens: yeah I hope so to, the delays in the getter functions don't seem to reliably work and I have no idea what is or isn't happening. I'm hopeful the jest tests will be better with the mocked timers, which should also result in a faster test suite.

src/components/builder/registration/registration-attribute.spec.tsx
sergei-maertens: I think the `input.focus()` also triggers an internal formik state update, perhaps only that needs to be wrapped in `act` :thinking: 

I'll see if I can move the `user.keyboard` out of the `act` when moving tests to jest.

PR #106, opened 2023-12-28
Nullable defaultValue for currency/number components

### LINKED ISSUE #105: Number + currency don't accept `null` for defaultValue (amongst others)
Formio 4.13 itself initializes `defaultValue: null` for components that don't define it explicitly, and this also makes sense (semantically) for Number/Currency components, since (empty) string values are a type mismatch.

https://github.com/formio/formio.js/blob/bebc2ad73cad138a6de0a8247df47f0085a314cc/src/components/_classes/component/Component.js#L78

`validate.min` and `validate.max` get even crazier:

https://github.com/formio/formio.js/blob/bebc2ad73cad138a6de0a8247df47f0085a314cc/src/components/number/Number.js#L14

sergei-maertens:
Closes #105

### DIFF
--- a/src/tests/EditorStub.ts
+++ b/src/tests/EditorStub.ts
@@ -1,0 +1,3 @@
+// Stub out CKEditor to prevent errors from its
+// CSS build polluting the test output.
+export default class EditorStub {}
--- a/package-lock.json
+++ b/package-lock.json
@@ -4922,9 +4922,9 @@
       }
     },
     "node_modules/@open-formulieren/types": {
-      "version": "0.18.2",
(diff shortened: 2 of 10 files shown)


src/tests/EditorStub.ts
Viicos, line 1: No idea what https://github.com/open-formulieren/formio-builder/pull/106/commits/8753f696baa628c3563524a2aa547aecaaadd94f is doing but apart from that looks good

Viicos (approved):

src/tests/EditorStub.ts
sergei-maertens: not doing it tries to load the real editor, which injects style tags and jsdom apparently fails to parse the minified CSS, cluttering the jest output massively.

PR #107, opened 2023-12-28
:children_crossing: Use floating-ui for tooltip implementation

### LINKED ISSUE #35: Properly implement tooltips

sergei-maertens:
Closes #35

This has better handling for overflow, accessibility and is more customizable. Additionally, it allows rendering any ReactNode instead of being limited to strings only.

The clipping/overflow behaviour is still a bit wonkey - because the modal itself applies `overflow: hidden` and that would require us to set up manual overflow checking. Considering the UI of the form designer, I don't expect any issues here and the tooltips behave correctly in the main `ComponentConfiguration` stories.

### DIFF
--- a/src/components/formio/tooltip.tsx
+++ b/src/components/formio/tooltip.tsx
@@ -32,7 +32,13 @@ const Tooltip: React.FC<TooltipProps> = ({text}) => {
     placement: 'right',
   });
 
-  const hover = useHover(context, {move: false});
+  const hover = useHover(context, {
+    move: false,
+    delay: {
+      open: 0,
+      close: 100,
+    },
+  });
   const focus = useFocus(context);
   const dismiss = useDismiss(context);
   const role = useRole(context, {role: 'label'});
--- a/src/components/formio/tooltip.tsx
+++ b/src/components/formio/tooltip.tsx
@@ -1,18 +1,72 @@
-import uniqueId from 'lodash.uniqueid';
-import React from 'react';
-import {Tooltip as ReactToolTip} from 'react-tooltip';
+/**
+ * Implements a tooltip using float-ui.
+ *
+ * Guide from https://floating-ui.com/docs/tooltip.
+ */
+import {
+  autoUpdate,
+  flip,
+  offset,
+  shift,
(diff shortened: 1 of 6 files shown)


src/components/formio/tooltip.tsx
```diff
@@ -1,18 +1,72 @@
...
+      {isOpen && (
+        <div
+          ref={refs.setFloating}
+          style={{
+            maxInlineSize: 'var(--of-builder-tooltip-max-inline-size, 200px)',
+            padding: 'var(--of-builder-tooltip-padding, 0.25rem 0.5rem)',
+            color: 'var(--of-builder-tooltip-color, #fff)',
+            textAlign: 'center',
+            backgroundColor: 'var(--of-builder-tooltip-background-color, #000)',
+            borderRadius: 'var(--of-builder-tooltip-border-radius, 0.25rem)',
```
sergei-maertens, line 61: These allow us to add proper dark-theme support at some point :)

Viicos (approved):
Any idea if we could leave the tooltip open when hovering it?

Currently, when hovering the tooltip icon and moving the mouse away, there's a small delay before the tooltip actually disappears. This way I'm able to hover the tooltip rectangle, and select text in it. If not possible this isn't a big deal

sergei-maertens:
> Any idea if we could leave the tooltip open when hovering it?

Done - was as simple as setting a 100ms delay for closing so that you have 100ms to move the cursor from reference element to tooltip.

PR #110, opened 2024-01-16
[open-formulieren/open-forms#3774] Fix dark mode compatibility

Viicos:
Part of https://github.com/open-formulieren/open-forms/issues/3774


vaszig (approved):
Let's see how this is going when merged.

PR #117, opened 2024-01-26
Support dark mode for the react-select widgets

sergei-maertens:
Partly fixes open-formulieren/open-forms#3774


Viicos (approved):
Was able to test in the django admin and can confirm it works

PR #124, opened 2024-02-12
Fix multiple option breaking form preview

sergei-maertens:
Discovered while investigating open-formulieren/open-forms#3858

`defaultValue: [null]` crashes the preview (rightfully so).

The backend/Formio adds the `defaultValue: null`, so when toggling multiple, this turns it into `[null]`. Setting it to `defaultValue: []` (always) results in `[[]]` which is also wrong.

The `Multiple` component should not be aware of special behaviour for particular component types, so instead we allow the component edit form to control whether it can touch the `defaultValue` or not.

### DIFF
--- a/src/components/builder/multiple.tsx
+++ b/src/components/builder/multiple.tsx
@@ -10,7 +10,17 @@ const hasDefaultValue = (component: any): component is AnyComponentSchema & {def
   return hasOwnProperty(component, 'defaultValue');
 };
 
-function Multiple<T = unknown>() {
+export interface MultipleProps {
+  /**
+   * If true, toggling the 'multiple' value will toggle the default value accordingly.
+   *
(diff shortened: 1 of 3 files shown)


src/components/builder/multiple.tsx
```diff
@@ -22,6 +26,10 @@ function Multiple<T = unknown>() {
 
   const onChange = (e: React.ChangeEvent<HTMLInputElement>) => {
     formikOnChange(e);
+
+    // only touch default value if we're allowed to
+    if (!manageDefaultValue) return;
```
Viicos, line 31: Maybe `alterDefaultValue`? Not satisfied either with this one, so maybe add a docstring in props to quickly describe what it does

Viicos (approved):

src/components/builder/multiple.tsx
sergei-maertens: went for `updateDefaultValue` + docstring

PR #125, opened 2024-02-27
:bug: [#3920] add isClearable to simpleConditional

nikkiysendoorn1:
Closes open-formulieren/open-forms#3920

I added the isClearable prop. This component doesn't have the prop required so, I think that this is enough. I thought maybe I need to look if it is required because then the `isClearable` should be to false.


sergei-maertens (changes requested):
We also need a similar change to the `content` component style dropdown, see the comment on the original issue

sergei-maertens (approved):

sergei-maertens:
Backport commit: 38bf0c2dff51a6fb5a2bc6751c03e492ab2835a4 (squashing both of the commits of this PR into one)

PR #127, opened 2024-02-27
Fix default/empty values for component prefill configuration

sergei-maertens:
Partly fixes open-formulieren/open-forms#3922

### DIFF
--- a/src/components/formio/select.tsx
+++ b/src/components/formio/select.tsx
@@ -7,11 +7,16 @@ import type {
   Props as RSProps,
   ThemeConfig,
 } from 'react-select/dist/declarations/src';
+import {MultiValue} from 'react-select/dist/declarations/src';
 
 import Component from './component';
 import Description from './description';
 
+// alias so that we can keep track of them and improve with generics at some point.
+type ValueType = any;
+
 // See https://react-select.com/typescript
+// TODO: revisit the generics here, they were from before I better understood TypeScript.
 
 export interface SelectProps<
   Option = unknown,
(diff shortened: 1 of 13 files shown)


src/components/formio/select.tsx
sergei-maertens, line 1: the typescript casting in this file is horrible... I remember dealing with react-select and their generics as one of the first typescript things I did and we definitely need to revisit this, but I'm putting that out of scope for this bugfix right now.

src/components/formio/select.tsx
```diff
@@ -202,10 +211,16 @@ function Select<
...
             const isSingle = !Array.isArray(newValue);
-            const normalized = isSingle ? [newValue] : newValue;
-            const rawValues = normalized.map(val => val?.[valueProperty] ?? null);
+
+            // for multiple, the value is reset to empty array (`[]`)
+            if (action.action === 'clear' && isSingle) {
+              newValue = emptyValue;
+            }
+
+            const normalized = (isSingle ? [newValue] : newValue) as MultiValue<Option>;
```
vaszig, line 222: Why do we explicitly declare the result type here?

vaszig (approved):

src/components/formio/select.tsx
sergei-maertens: otherwise the typescript compiler complains about implicit `any` types :)

sergei-maertens:
Backport: a74a2d354dc769bbe168bf35c2c0d2fe089892f5

PR #129, opened 2024-02-29
Include pointeers to templates documentation in the builder

sergei-maertens:
Not all fields are covered yet, but placeholder only supports this for *some* components.

At least the label, tooltip, description and content HTML fields now inform the form designer that they can use Django template syntax.


vaszig (changes requested):
A few small remarks for which I may have understood something wrong but yeah,please let me know if so.

PR #130, opened 2024-03-06
[open-formulieren/open-forms#3943] Fix validation of number decimal places

Viicos:
Fixes open-formulieren/open-forms#3943.

Did the same as the currency component (this this comment: https://github.com/open-formulieren/formio-builder/pull/50#discussion_r1386164089), but then figured out a better solution


vaszig (approved):
This was fast and clear!thanks

PR #131, opened 2024-03-06
:sparkles: [#483] add description for radio en checkbox option

nikkiysendoorn1:
Closes open-formulieren/open-forms-sdk#483

### DIFF
--- a/src/components/builder/values/option-row.tsx
+++ b/src/components/builder/values/option-row.tsx
@@ -6,10 +6,12 @@ import {useState} from 'react';
 import {FormattedMessage, useIntl} from 'react-intl';
 
 import {TextField} from '@/components/formio';
+import ComponentLabel from '@/components/formio/component-label';
 import {TextArea} from '@/components/formio/textarea';
 
 const ICONS_CELL = css`
   vertical-align: middle !important;
+  border-bottom: none !important;
 `;
 
 const SORT_ICONS = css`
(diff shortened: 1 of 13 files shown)


Viicos (commented):
If I remember correctly, the issue was to add a description for checkboxes in the SDK (that is, available to the end user). In the form builder (this repo), we actually need to add a text input for each option where the description, so probably a 3rd column here:

image

? Maybe we should confirm with @joeribekker

sergei-maertens (commented):
Couple thoughts/remarks:

* the description takes up too much space. Rather, make this "second row" collapsed by default, and add a link/icon to "edit option description" which expands/collapses it on click
* I can't seem to enter translations yet for these descriptions :)

src/components/builder/values/option-row.tsx
```diff
@@ -63,69 +64,87 @@ const OptionRow: React.FC<OptionRowProps> = ({name, index, arrayHelpers}) => {
...
+          <FormattedMessage
+            description="Option label table header/label"
+            defaultMessage="Description"
+          />
+          <TextArea
+            name={`${fieldNamePrefix}.description`}
+            aria-label={intl.formatMessage({
+              description: 'Accessible label for option label',
+              defaultMessage: 'Option description',
+            })}
```
sergei-maertens, line 144:
```suggestion
            aria-label={intl.formatMessage({
              description: 'Accessible label for option description',
              defaultMessage: 'Option description',
            })}
```

sergei-maertens (changes requested):
What I meant was that when I enter the description of an option, there's no way to go to the "translations" tab and enter the Dutch/English translation for that same description.

If a form is being filled out in English, the option labels (and descriptions) must display their english translations instead of the default/Dutch translations. Currently the person defining the form cannot enter these translations anywhere.

sergei-maertens (approved):
I've done some more UI tweaks :)

PR #135, opened 2024-04-04
[open-formulieren/open-forms#4084] Fix default value of multiple `select` component

Viicos:
Partly fixes open-formulieren/open-forms#4084


sergei-maertens (changes requested):
Regression test please :)

sergei-maertens (approved):

PR #136, opened 2024-04-05
[OF#3964] Make input field in 'simple conditional' dependent on component type

SilviaAmAm:
Fixes https://github.com/open-formulieren/open-forms/issues/3964

Todo:
- [x] Fix currency component
- [ ] See if you can fix bool component and remove backend fix => Should maybe go in its own ticket and other components could be revisited too? :thinking: 
- [x] add stories/tests


src/components/builder/comparison-value-input.tsx
```diff
@@ -0,0 +1,60 @@
+import {useFormikContext} from 'formik';
+import {ExtendedComponentSchema, Utils as FormioUtils} from 'formiojs';
+import {useContext, useEffect, useState} from 'react';
+import usePrevious from 'react-use/esm/usePrevious';
+
+import {BuilderContext} from '@/context';
+import {getRegistryEntry} from '@/registry';
```
sergei-maertens, line 7: I suspect this import is causing the circular import, leading to items being `undefined`

src/components/builder/comparison-value-input.tsx
```diff
@@ -0,0 +1,60 @@
...
+    const component: ExtendedComponentSchema = FormioUtils.getComponent(
+      getFormComponents(),
+      conditional.when,
+      false
+    );
+
+    setChosenComponent(component);
+  }, [values, setFieldValue]);
+
+  if (!chosenComponent) return null;
```
Viicos, line 47: Probably this doesn't create issues in our case, but it seems to be better [to return `false` instead](https://medium.com/@davidkelley87/stop-using-return-null-in-react-a2ebf08fc9cd)

src/components/builder/comparison-value-input.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+  const registryEntry = getRegistryEntry(chosenComponent);
+  const {comparisonValueComponent} = registryEntry;
+  const ComparisonValueInputComponent =
+    comparisonValueComponent || Fallback.comparisonValueComponent;
```
Viicos, line 52: Maybe something like

```suggestion
  const {comparisonValueComponent = Fallback.comparisonValueComponent} = registryEntry;
```

Or maybe `registryEntry.comparisonValueComponent` can be `null` so in that case your solution is safer

src/components/builder/simple-conditional.stories.tsx
```diff
@@ -54,3 +56,82 @@ const Template: StoryFn<typeof SimpleConditional> = () => <SimpleConditional />;
...
+
+    eqInput = canvas.queryByLabelText('Has the value');
+
+    expect(eqInput).not.toBeNull();
+    expect(eqInput.type).toEqual('text');
+
+    const svgTags = document.getElementsByTagName('svg');
+    // 0 is the cross for the 'This component should display' field, 1 is the arrow of the dropdown
+    const littleCross = svgTags[2];
+    await userEvent.click(littleCross);
```
Viicos, line 87: I don't know if it's possible, but a class selector might more robust?

Viicos (approved):

src/components/builder/simple-conditional.stories.tsx
sergei-maertens: react-select is notoriously bad for that :/

src/components/builder/comparison-value-input.tsx
```diff
@@ -0,0 +1,60 @@
...
+import {getRegistryEntry} from '@/registry';
+import Fallback from '@/registry/fallback';
+import {comparisonValueComponentProps} from '@/registry/types';
+
+export interface simpleConditional {
+  conditional: {
+    eq: string;
+    when: string;
+    show: null | boolean;
+  };
```
sergei-maertens, line 16: this doesn't exactly line up with the formio type definition: https://github.com/formio/formio.js/blob/bebc2ad73cad138a6de0a8247df47f0085a314cc/types/components/schema.d.ts#L157, which we rely on (semi-implicitly) via https://github.com/open-formulieren/types/blob/5a83f493525aeeb9ca5bca6386429853479a6536/src/formio/base.ts#L106

Most notably:

* `show` is not allowed to be `null`, it needs to be `bool` or undefined
* `json` is not specified, but since we don't make use of it, okay to leave this out.

I'd probably define this as:

```ts
import type { ConditionalOptions } from 'formiojs';

export interface SimpleConditional {
  conditional: Omit<ConditionalOptions, 'json'>;
}
```
to avoid type checker issues in the future.

I'm also not sure about making `conditional` a required property - the upstream type has it as optional: https://github.com/formio/formio.js/blob/bebc2ad73cad138a6de0a8247df47f0085a314cc/types/components/schema.d.ts#L87

src/components/builder/comparison-value-input.tsx
```diff
@@ -0,0 +1,60 @@
...
+}
+
+export const ComparisonValueInput: React.FC = () => {
+  const {getFormComponents} = useContext(BuilderContext);
+  const {values, setFieldValue} = useFormikContext<simpleConditional>();
+
+  const [chosenComponent, setChosenComponent] = useState<ExtendedComponentSchema | null>(null);
```
sergei-maertens, line 23: is it really necessary to track extra state for something that can easily be computed during render? I prefer to not do that since keeping state in sync is not ideal. Instead I'd rather see something like:

```ts
  const componentKey = values.conditional?.when;
  const chosenComponent: ExtendedComponentSchema = componentKey 
    ? FormioUtils.getComponent(getFormComponents(), conditional.when, false)
    : null;

  useEffect(...);

  if (chosenComponent === null) return false;

  // rest of render logic
```

src/components/builder/comparison-value-input.tsx
sergei-maertens, line 52: I think we don't need this layer of going through the registry at all, instead, why don't we pass the `Component` as a prop in the `EditForm`?

E.g. for number/currency:

```tsx
const EditForm = () => {
  // hooks and stuff
  return (
    <Tabs>
      <TabList />

      {/* Advanced tab */}
      <TabPanel>
        <SimpleConditional valueAs={NumberField} />
      </TabPanel>

      {/* ... rest of tab content */}
    </Tabs>
  );
};
```

so then you can easily keep the default value of `valueAs` prop to the `TextField`, which keeps it backwards compatible, and you update the currency/number edit forms.

Advantage is that we don't have to expand the interface of the `RegistryEntry` (particularly with optional properties - with TS I prefer to make as much as possible mandatory because it simplifies reasoning about things), and we keep the same drawback: you have to remember to specify the correct field type for the value to be rendered.

Alternatively, instead of passing a component "type", you could also accept a render function (which is more akin to best practices):

```tsx
<SimpleConditional renderValue={valueProps => <NumberField {...valueProps} />} />
```
which in itself is nice because you can manage the `multiple` at this point instead of generically, so I think this would have my preference even :thinking:

src/registry/currency/edit.tsx
```diff
@@ -159,6 +159,20 @@ const DefaultValue: React.FC = () => {
   );
 };
 
+const ComparisonValueInput: React.FC = () => {
+  return (
+    <NumberField
+      name="conditional.eq"
```
sergei-maertens, line 165: with my above suggestions, we don't need to duplicate the form field name and literalsf or label/tooltip anymore :)

sergei-maertens (changes requested):

src/components/builder/comparison-value-input.tsx
SilviaAmAm: I don't think we can pass the component in the edit form as prop? Maybe I am misunderstanding what you mean.

In the edit form, we don't know in advance which component will be selected in the condition. => We use the type of the component which is selected in the conditional tab, not the type of the component corresponding to the EditForm we are in.

sergei-maertens: Ah yes, I was confusing the type of the component we're setting up with the constraint for the eq value, but that doesn't hold up, since you need the data type of the reference component. And the same applies for the `multiple` aspect.

Multiple I find tricky in general because File field for example has different `multiple` behaviour, but that's out of scope for here now.

Keep it as is, I was wrong!

src/components/builder/comparison-value-input.tsx
SilviaAmAm: After further discussion, we decided to stick to `null` since the react docs say to use null: https://react.dev/learn/conditional-rendering#conditionally-returning-nothing-with-null

Viicos: I don't know who to trust anymore in this frontend world :shrug: :shrug: :shrug: :shrug:

sergei-maertens: The `React.FC` type declarations are a good starting point :P they don't allow `false` as return value

Viicos: Ok well reading the comments of the article seems to be a good starting point as well

sergei-maertens (approved):

PR #138, opened 2024-04-10
[open-formulieren/open-forms#4084] Handle `select`s `defaultValue` when set to `null`

Viicos:
Partly fixes open-formulieren/open-forms#4084


src/registry/select/select.spec.tsx
```diff
@@ -62,6 +62,57 @@ test('Switch to multiple sets empty array as default value', async () => {
...
+    label: 'A select field',
+    dataSrc: 'values',
+    openForms: {
+      dataSrc: 'manual',
+      translations: {},
+    },
+    data: {
+      values: [{value: 'opt1', label: 'Opt1'}],
+    },
+    // @ts-ignore
```
sergei-maertens, line 82:
```suggestion
    // @ts-expect-error
```
I think this is more appropriate, and I think it'll also give a notice when TS does not error :)

sergei-maertens (approved):

src/registry/select/select.spec.tsx
Viicos: Interesting, I read about it https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-9.html#ts-ignore-or-ts-expect-error. Although I think I prefer the Python way: using `type: ignore` for everything, and enable the reporting of unnecessary comments in mypy/pyright ;)

PR #141, opened 2024-04-12
Setup sass toolchain

sergei-maertens:
To replace the `@emotion/css` usage, which injects styles from a CSS-in-JS solution.

Using sass allows us to extract the styles into a single stylesheet that can be loaded separately, while leveraging the pre-processing/reusability from SASS. We can use predicatable (and overrideable) class names for the components & benefit from proper syntax highlighting.

- [x] Get imports working with `tsc` compilation
- [x] Get SCSS working with Storybook
- [x] Get CI to pass
- [x] Set up sass bundling for production/package distribution
- [x] Refactor remaining `@emotion/css` usage


src/registry/editgrid/preview.tsx
```diff
@@ -43,27 +22,27 @@ const Preview: React.FC<ComponentPreviewProps<EditGridComponentSchema>> = ({comp
       tooltip={tooltip}
     >
       <div style={{display: 'flex', flexDirection: 'column', gap: '1em'}}>
```
vaszig, line 24: I am a bit confused..Shouldn't we put all the css in a separate file now that we got rid of `@emotion/css`

vaszig (approved):
Why didn't we go from the beginning with pure css files and we were using the library?

sergei-maertens:
> Why didn't we go from the beginning with pure css files and we were using the library?

Because I didn't have the time then to figure out a proper working setup and hoped we could just get by with using bootstrap class names. Eventually it turned out that we needed some extra things but setting up a toolchain/packaging solution is quite complex :(

src/registry/editgrid/preview.tsx
sergei-maertens: inline `style` props do not use `@emotion/css` at all - but yeah, they will need to be converted/cleaned up too at some point.

My main goal here is being able to drop the CSS-in-JS dependency and being able to use SASS syntax.

PR #148, opened 2024-04-17
[OF#3964,#2900] Special simple conditional for checkboxes

SilviaAmAm:
Fixes https://github.com/open-formulieren/open-forms/issues/2900, https://github.com/open-formulieren/open-forms/issues/3964 (partially)

### DIFF
--- a/src/registry/checkbox/edit.tsx
+++ b/src/registry/checkbox/edit.tsx
@@ -19,10 +19,10 @@ import {
   useDeriveComponentKey,
 } from '@/components/builder';
 import {LABELS} from '@/components/builder/messages';
-import {Checkbox, TabList, TabPanel, Tabs} from '@/components/formio';
+import {Checkbox, Select, TabList, TabPanel, Tabs} from '@/components/formio';
 import {useErrorChecker} from '@/utils/errors';
 
-import {EditFormDefinition} from '../types';
+import {ComparisonValueProps, EditFormDefinition} from '../types';
 
 /**
  * Form to configure a Formio 'checkbox' type component.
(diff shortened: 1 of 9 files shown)


src/registry/checkbox/edit.tsx
```diff
@@ -155,4 +155,33 @@ const DefaultValue: React.FC<DefaultValueProps> = ({description}) => {
...
+              description="Label of the 'false' checkbox option"
+              defaultMessage="Not checked"
+            />
+          ),
+        },
+      ]}
+    />
+  );
+};
+
```
Viicos, line 185: Would a checkbox component be better suited? Or did you have to use a select as the `Has the value` label is not customizable (well not anymore with your other changes), so it would look weird with a checkbox

Viicos (approved):

src/registry/checkbox/edit.tsx
SilviaAmAm: I thought that it looked a bit weird the 'has the value' with the checkbox, that's why I went for the dropdown :thinking:

src/registry/checkbox/edit.tsx
```diff
@@ -155,4 +155,33 @@ const DefaultValue: React.FC<DefaultValueProps> = ({description}) => {
   );
 };
 
+const renderComparisonValueInput: React.FC<ComparisonValueProps> = props => {
```
sergei-maertens, line 158: Hmm, this is still just a regular React component (they're just functions, basically), so the "restructuring" to render functions here doesn't seem to make a lot of sense?

```tsx
const ComparisonValueInput: React.FC<ComparisonValueProps> = props => (
  <Select {...props} options={...} />
);
```

would be 100% equivalent, no?

src/registry/currency/edit.tsx
```diff
@@ -159,18 +159,8 @@ const DefaultValue: React.FC = () => {
...
-      name="conditional.eq"
-      label={
-        <FormattedMessage
-          description="Component property 'conditional.eq' label"
-          defaultMessage="Has the value"
-        />
-      }
-    />
-  );
+const renderComparisonValueInput: React.FC<ComparisonValueProps> = props => {
```
sergei-maertens, line 162: what about:

```tsx
const ComparisonValueInput = NumberField;
```

All the `ComparisonValueProps` are/should be assignable to `NumberField`, so I'd expect this to type-check.

If it doesn't, a higher order component could work:

```tsx
const ComparisonValueInput = asComparisonValue(NumberField);
```

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #150, opened 2024-04-18
 [open-formulieren/open-forms#4158] Add new error keys for date(time) 

Viicos:
Part of open-formulieren/open-forms#4158


vaszig (approved):
I guess the errors and the CI failure have to do with the types repository you updated as well, right?

sergei-maertens (approved):

PR #151, opened 2024-04-24
[OF#4064] Ability to remove default value from radio

vaszig:
Fixes open-formulieren/open-forms#4064

### DIFF
--- a/src/components/formio/radio.stories.ts
+++ b/src/components/formio/radio.stories.ts
@@ -110,13 +110,13 @@ export const DefaultValueRemoved: Story = {
     const input2 = canvas.getByLabelText('B');
 
     // Before clearing the default value
-    await expect(canvas.queryByText('Clear radio button values')).toBeInTheDocument();
+    await expect(canvas.queryByText('Clear selection')).toBeInTheDocument();
     await expect(input1).toBeChecked();
     await expect(input2).not.toBeChecked();
 
     // After clearing the default value
-    await userEvent.click(canvas.getByRole('button', {name: 'Clear radio button values'}));
-    await expect(canvas.queryByText('Clear radio button values')).not.toBeInTheDocument();
+    await userEvent.click(canvas.getByRole('button', {name: 'Clear selection'}));
+    await expect(canvas.queryByText('Clear selection')).not.toBeInTheDocument();
     await expect(input1).not.toBeChecked();
     await expect(input2).not.toBeChecked();
   },
--- a/src/components/formio/radio.stories.ts
+++ b/src/components/formio/radio.stories.ts
@@ -91,7 +91,8 @@ export const WithErrors: Story = {
 
 export const DefaultValueRemoved: Story = {
   args: {
-    label: 'With removing the default value',
+    label: 'Clearable default value',
+    isClearable: true,
     options: [
       {value: 'a', label: 'A'},
       {value: 'b', label: 'B'},
@@ -106,17 +107,18 @@ export const DefaultValueRemoved: Story = {
 
   play: async ({canvasElement}) => {
     const canvas = within(canvasElement);
+    const button = canvas.getByRole('button', {name: 'Clear selection'});
     const input1 = canvas.getByLabelText('A');
     const input2 = canvas.getByLabelText('B');
 
     // Before clearing the default value
-    await expect(canvas.queryByText('Clear selection')).toBeInTheDocument();
+    expect(button).toBeVisible();
     await expect(input1).toBeChecked();
     await expect(input2).not.toBeChecked();
 
     // After clearing the default value
-    await userEvent.click(canvas.getByRole('button', {name: 'Clear selection'}));
-    await expect(canvas.queryByText('Clear selection')).not.toBeInTheDocument();
+    await userEvent.click(button);
+    expect(button).not.toBeVisible();
     await expect(input1).not.toBeChecked();
     await expect(input2).not.toBeChecked();
   },
--- a/src/components/formio/radio.stories.ts
+++ b/src/components/formio/radio.stories.ts
@@ -89,9 +89,9 @@ export const WithErrors: Story = {
   },
 };
 
-export const DefaultValueRemoved: Story = {
+export const Clearable: Story = {
   args: {
-    label: 'Clearable default value',
+    label: 'Clearable',
     isClearable: true,
     options: [
       {value: 'a', label: 'A'},
@@ -121,5 +121,21 @@ export const DefaultValueRemoved: Story = {
     expect(button).not.toBeVisible();
     await expect(input1).not.toBeChecked();
     await expect(input2).not.toBeChecked();
+  },
+};
+
+export const ClearableWithEmptyStringOption: Story = {
+  args: {
+    label: 'Clearable',
+    isClearable: true,
+    options: [
+      {value: '', label: ''}, // rendering is weird with empty labels
+    ],
+  },
+
+  parameters: {
+    formik: {
+      initialValues: {'my-radio': ''},
+    },
   },
 };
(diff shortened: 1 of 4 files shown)


src/components/formio/radio.tsx
```diff
@@ -51,11 +54,36 @@ export const Radio: React.FC<RadioProps> = ({
   options,
   label,
   required = false,
+  isPreview = false,
   tooltip = '',
   description = '',
 }) => {
+  const {values, setValues} = useFormikContext<ExtendedComponentSchema>();
```
Viicos, line 61:
```suggestion
  const {values, setFieldValue} = useFormikContext<ExtendedComponentSchema>();
```

and then `onClick={() => setFieldValue(name, '')}`

src/components/formio/radio.tsx
```diff
@@ -51,11 +54,36 @@ export const Radio: React.FC<RadioProps> = ({
...
+        <div>
+          <button
+            type="button"
+            className="btn btn-secondary formio-button-remove-row"
+            onClick={clearRadioValues}
+          >
+            <i className="fa fa-times-circle" aria-hidden="true" />
+            <FormattedMessage
+              description="Clear radio values button label"
+              defaultMessage="Clear radio button values"
```
Viicos, line 81: "Clear selection" is probably shorter.

I think it's also better to put this button under the radio element, so that it doesn't move the radios when one gets selected

src/registry/radio/preview.tsx
```diff
@@ -42,6 +42,7 @@ const Preview: React.FC<ComponentPreviewProps<RadioComponentSchema>> = ({compone
       label={label}
       tooltip={tooltip}
       required={required}
+      isPreview={true}
```
Viicos, line 45: I'll go with `clearable`, so that it is not related to the preview functionality we have in the builder

Viicos (changes requested):
image

The button icon probably requires some padding/margin

src/components/formio/radio.stories.ts
```diff
@@ -88,3 +88,36 @@ export const WithErrors: Story = {
     await expect(canvas.queryByText('Example error')).toBeInTheDocument();
   },
 };
+
+export const DefaultValueRemoved: Story = {
+  args: {
+    label: 'With removing the default value',
```
sergei-maertens, line 94:
```suggestion
    label: 'Clearable default value',
```

src/components/formio/radio.stories.ts
```diff
@@ -88,3 +88,36 @@ export const WithErrors: Story = {
...
+    },
+  },
+
+  play: async ({canvasElement}) => {
+    const canvas = within(canvasElement);
+    const input1 = canvas.getByLabelText('A');
+    const input2 = canvas.getByLabelText('B');
+
+    // Before clearing the default value
+    await expect(canvas.queryByText('Clear selection')).toBeInTheDocument();
```
sergei-maertens, line 113: can't this be the same `getByRole('button')` query? it's more specific than just finding the text in the document and you use it anyway to interact with it

src/components/formio/radio.stories.ts
```diff
@@ -88,3 +88,36 @@ export const WithErrors: Story = {
...
+    await expect(input1).toBeChecked();
+    await expect(input2).not.toBeChecked();
+
+    // After clearing the default value
+    await userEvent.click(canvas.getByRole('button', {name: 'Clear selection'}));
+    await expect(canvas.queryByText('Clear selection')).not.toBeInTheDocument();
```
sergei-maertens, line 119: perhaps you can re-use the button reference from earlier when you assert that it's present, that way the test is more robust in case the literal text changes:

```tsx
const button = canvas.getByRole('button', {name: 'Clear selection'});
expect(button).toBeVisible();
await userEvent.click(button);
expect(button).not.toBeInTheDocument();
```

I'm not sure this works though, but please give it a try.

src/components/formio/radio.stories.ts
```diff
@@ -88,3 +88,36 @@ export const WithErrors: Story = {
     await expect(canvas.queryByText('Example error')).toBeInTheDocument();
   },
 };
+
+export const DefaultValueRemoved: Story = {
+  args: {
```
sergei-maertens, line 93: I don't see the `isClearable` prop being passed that should enable this behaviour?

src/components/formio/radio.tsx
```diff
@@ -51,9 +55,16 @@ export const Radio: React.FC<RadioProps> = ({
   options,
   label,
   required = false,
+  isClearable = true,
```
sergei-maertens, line 58: let's default this to `false`. Idiomatic react is passing booleans as `<Radio isClearable />` which enables it, and not specifying it disables it.

Furthermore, I don't think default clearable makes sense, otherwise the browsers would've built in such behaviour natively for radio buttons :)

src/components/formio/radio.tsx
```diff
@@ -51,9 +55,16 @@ export const Radio: React.FC<RadioProps> = ({
...
   tooltip = '',
   description = '',
 }) => {
+  const {values, setFieldValue} = useFormikContext<ExtendedComponentSchema>();
+
+  const CLEAR_VALUES_ICON = css`
+    margin-right: 5px;
+  `;
```
sergei-maertens, line 66: this should be moved outside of the component implementation, it's a constant and doesn't need re-calculating on every render.

It will also cause merge conflicts with my #141  PR :(

src/components/formio/radio.tsx
```diff
@@ -67,6 +78,22 @@ export const Radio: React.FC<RadioProps> = ({
       </div>
 
       {description && <Description text={description} />}
+
+      {values[name] && isClearable && (
```
sergei-maertens, line 82: `values[name]` is too naive because dotted paths can be used which lead to nested lookups.

Use:

```tsx
const {getFieldProps} = useFormikContext();
const {value} = getFieldProps(name);
```
instead

sergei-maertens (changes requested):

src/registry/radio/edit.tsx
```diff
@@ -182,6 +182,7 @@ const DefaultValue: React.FC<DefaultValueProps> = ({options}) => {
       options={options}
       label={<FormattedMessage {...LABELS.defaultValue} />}
       tooltip={tooltip}
+      isClearable={true}
```
sergei-maertens, line 185:
```suggestion
      isClearable
```
this is more idiomatic React

src/components/formio/radio.tsx
```diff
@@ -1,6 +1,9 @@
+import {css} from '@emotion/css';
```
sergei-maertens, line 1: unused import - you'll also need to import the SCSS module once the sass-toolchain PR is merged

src/components/formio/radio.tsx
```diff
@@ -67,6 +75,22 @@ export const Radio: React.FC<RadioProps> = ({
...
+          <button
+            type="button"
+            className="btn btn-secondary formio-button-remove-row"
+            onClick={() => setFieldValue(name, '')}
+          >
+            <i className={clsx('fa', 'fa-times-circle', 'offb-radio-default')} aria-hidden="true" />
+            <FormattedMessage
+              description="Clear selection button label"
+              defaultMessage="Clear selection"
+            />
```
sergei-maertens, line 90: How does the spacing look if you just add some whitespace to the generated markup?

```suggestion
            <i className={clsx('fa', 'fa-times-circle', 'offb-radio-default')} aria-hidden="true" />
            {' '}
            <FormattedMessage
              description="Clear selection button label"
              defaultMessage="Clear selection"
            />
```
if this looks okay, then just remove the `offb-radio-default` class name and SCSS file.

sergei-maertens (changes requested):

src/components/formio/radio.tsx
vaszig: It works fine!I wouldn't have thought of this simple solution

sergei-maertens (approved):

PR #155, opened 2024-05-27
[OF#3993] Add deriveAddress boolean to addressNL component

vaszig:
Partly fixes open-formulieren/open-forms#3993

- [x] Add deriveAddress to the addressNL component
- [x] Install the new version of the types

### DIFF
--- a/src/registry/addressNL/edit.tsx
+++ b/src/registry/addressNL/edit.tsx
@@ -1,5 +1,5 @@
 import {AddressNLComponentSchema} from '@open-formulieren/types';
-import {useIntl} from 'react-intl';
+import {FormattedMessage, useIntl} from 'react-intl';
 
 import {
   BuilderTabs,
@@ -18,10 +18,32 @@ import {
   useDeriveComponentKey,
 } from '@/components/builder';
 import {LABELS} from '@/components/builder/messages';
+import {Checkbox} from '@/components/formio';
 import {TabList, TabPanel, Tabs} from '@/components/formio';
 import {useErrorChecker} from '@/utils/errors';
 
 import {EditFormDefinition} from '../types';
+
+const DeriveAddress = () => {
+  const intl = useIntl();
+  const tooltip = intl.formatMessage({
+    description: "Tooltip for 'DeriveAddress' builder field",
+    defaultMessage:
+      'When enabled, the street name and city are derived from the entered postcode and house number.',
+  });
+  return (
+    <Checkbox
+      name="deriveAddress"
+      label={
+        <FormattedMessage
+          description="Label for 'DeriveAddress' builder field"
+          defaultMessage="Derive address"
+        />
+      }
+      tooltip={tooltip}
+    />
+  );
+};
 
 /**
  * Form to configure a Formio 'address' type component.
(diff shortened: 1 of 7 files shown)


vaszig:
Requires the new version of types before merging. See open-formulieren/types#48 .

src/components/builder/derive-address.tsx
sergei-maertens, line 1: I think I'd define this component directly in `src/registry/addressNL/edit.tsx` and don't export it - it's only used in one place. This directory is meant for components/fields that are used in more formio components, such as label/description/required... 

There's also no need for individual storybook stories for this field/component, it's just a pre-configured checkbox :)

src/components/builder/derive-address.tsx
```diff
@@ -0,0 +1,25 @@
+import {FormattedMessage, useIntl} from 'react-intl';
+
+import {Checkbox} from '../formio';
+
+const DeriveAddress = () => {
+  const intl = useIntl();
+  const tooltip = intl.formatMessage({
+    description: "Tooltip for 'DeriveAddress' builder field",
+    defaultMessage: 'Derive address.',
```
sergei-maertens, line 9: You can be a bit more inspiring here :wink: 

```suggestion
    defaultMessage: 'When enabled, the street name and city are derived from the entered postcode and house number.',
```

be informative, provide useful context, don't just repeat the label, otherwise we can just leave out the tooltip entirely.

src/components/builder/derive-address.tsx
```diff
@@ -0,0 +1,25 @@
...
+  });
+  return (
+    <Checkbox
+      name="disabled"
```
sergei-maertens, line 13: that doesn't seem right, the checkbox controls the value of the `component.deriveAddress` property, so:

```suggestion
      name="deriveAddress"
```

src/registry/addressNL/edit.tsx
```diff
@@ -117,11 +119,15 @@ EditForm.defaultValues = {
...
   isSensitiveData: true,
+  deriveAddress: false,
   defaultValue: {
     postcode: '',
     houseNumber: '',
     houseLetter: '',
     houseNumberAddition: '',
+    city: '',
+    streetName: '',
+    secretStreetCity: '',
```
sergei-maertens, line 130: I think we can leave these out if we update the types PR

src/registry/addressNL/preview.tsx
```diff
@@ -56,6 +56,26 @@ const Preview: React.FC<ComponentPreviewProps<AddressNLComponentSchema>> = ({com
...
+      <TextField
+        name={`${key}.streetNumber`}
+        label={
+          <FormattedMessage
+            description="Label for addressNL streetName read only result"
+            defaultMessage="Street name"
+          />
+        }
+        disabled
+      />
```
sergei-maertens, line 78: These fields should only be displayed when `deriveAddress` is `true`, otherwise not. That applies for both the form builder and the SDK :)

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #157, opened 2024-06-13
Remove support for defaultValue translations

sergei-maertens:
Partly closes open-formulieren/open-forms#4362
Depends on https://github.com/open-formulieren/types/pull/49

It is unused and too complex with 'multiple: true' cases, causing crashes in the admin interface.


vaszig (approved):
Approved since it will be happy after updating types version.

PR #160, opened 2024-06-19
Deprecate textfield location derivation

sergei-maertens:
Update the builder to steer people to the new component instead of using the legacy configuration.


vaszig (approved):
I missed that part, thanks.

PR #167, opened 2024-07-08
:sparkles: [#4420] Validation for fields within AddressNL component

stevenbal:
partially closes https://github.com/open-formulieren/open-forms/issues/4420

~~Requires https://github.com/open-formulieren/types/pull/51 to be merged and released first~~

### DIFF
--- a/src/components/builder/validate/i18n.tsx
+++ b/src/components/builder/validate/i18n.tsx
@@ -9,12 +9,12 @@ import {BuilderContext} from '@/context';
 import {DataMap, Panel, Tab, TabList, TabPanel, Tabs, TextField} from '../../formio';
 
 export function useManageValidatorsTranslations<S extends SchemaWithValidation>(
-  keys: PossibleValidatorErrorKeys<S>[]
+  keys: PossibleValidatorErrorKeys<S>[],
+  field: string = 'translatedErrors'
 ): void {
   const {supportedLanguageCodes} = useContext(BuilderContext);
-  const [{value}, , {setValue}] = useField<S['translatedErrors']>('translatedErrors');
+  const [{value}, , {setValue}] = useField<S['translatedErrors']>(field);
 
-  // set any missing translations
   useEffect(() => {
     const newValue = value
       ? {...value}
--- a/src/components/builder/validate/i18n.tsx
+++ b/src/components/builder/validate/i18n.tsx
@@ -2,7 +2,7 @@ import {PossibleValidatorErrorKeys, SchemaWithValidation} from '@open-formuliere
 import {useField} from 'formik';
 import {isEqual} from 'lodash';
 import {useContext, useEffect} from 'react';
-import {FormattedMessage, useIntl} from 'react-intl';
+import {FormattedMessage, defineMessage, useIntl} from 'react-intl';
 
 import {BuilderContext} from '@/context';
 
@@ -10,10 +10,11 @@ import {DataMap, Panel, Tab, TabList, TabPanel, Tabs, TextField} from '../../for
 
 export function useManageValidatorsTranslations<S extends SchemaWithValidation>(
   keys: PossibleValidatorErrorKeys<S>[],
-  field: string = 'translatedErrors'
+  prefix: string = ''
 ): void {
+  const fieldName = `${prefix}${prefix ? '.' : ''}translatedErrors`;
   const {supportedLanguageCodes} = useContext(BuilderContext);
-  const [{value}, , {setValue}] = useField<S['translatedErrors']>(field);
+  const [{value}, , {setValue}] = useField<S['translatedErrors']>(fieldName);
 
   useEffect(() => {
     const newValue = value
@@ -68,6 +69,10 @@ const ValidationErrorTranslations = () => {
                   defaultMessage="Error code"
                 />
               }
+              ariaLabelMessage={defineMessage({
+                description: 'Accessible label for error message input field',
+                defaultMessage: 'Error message for "{key}"',
+              })}
               valueComponent={
                 <TextField
                   name="message"
--- a/src/components/formio/datamap.tsx
+++ b/src/components/formio/datamap.tsx
@@ -1,5 +1,6 @@
 import {useField} from 'formik';
 import React from 'react';
+import {MessageDescriptor, useIntl} from 'react-intl';
 
 import DataGrid, {DataGridRow} from './datagrid';
 import TextField from './textfield';
@@ -8,12 +9,19 @@ export interface DataMapProps {
   name: string;
   valueComponent: React.ReactElement;
   keyLabel: React.ReactNode;
+  ariaLabelMessage?: MessageDescriptor;
 }
 
 /**
  * Data map with readonly keys.
  */
-export const DataMap: React.FC<DataMapProps> = ({name, valueComponent, keyLabel = 'Key'}) => {
+export const DataMap: React.FC<DataMapProps> = ({
+  name,
+  valueComponent,
+  ariaLabelMessage,
+  keyLabel = 'Key',
+}) => {
+  const intl = useIntl();
   const [{value}, , {setValue}] = useField(name);
   const transformedValue = Object.entries(value).map(([key, value]) => ({key, value}));
   const columns = [keyLabel, valueComponent.props.label];
@@ -30,6 +38,9 @@ export const DataMap: React.FC<DataMapProps> = ({name, valueComponent, keyLabel 
               const newValue = {...value, [item.key]: event.target.value};
               setValue(newValue);
             },
+            'aria-label': ariaLabelMessage
+              ? intl.formatMessage(ariaLabelMessage, {key: item.key})
+              : undefined,
           })}
         </DataGridRow>
       ))}
--- a/src/registry/addressNL/edit.stories.ts
+++ b/src/registry/addressNL/edit.stories.ts
@@ -1,5 +1,5 @@
 import {Meta, StoryObj} from '@storybook/react';
-import {expect, userEvent, waitFor, within} from '@storybook/test';
+import {expect, userEvent, within} from '@storybook/test';
 
 import ComponentEditForm from '@/components/ComponentEditForm';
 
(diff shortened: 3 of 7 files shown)


src/registry/addressNL/edit.tsx
```diff
@@ -19,12 +21,71 @@ import {
...
-import {TabList, TabPanel, Tabs} from '@/components/formio';
+import {DataMap, Panel, Tab, TabList, TabPanel, Tabs} from '@/components/formio';
 import {Select} from '@/components/formio';
+import {BuilderContext} from '@/context';
 import {useErrorChecker} from '@/utils/errors';
 
 import {EditFormDefinition} from '../types';
 
+export interface SubcomponentValidationProps {
+  component: string;
```
sergei-maertens, line 32:
```suggestion
  component: 'postcode' | 'city'; // or even `keyof Required<AddressNLComponentSchema>['openForms']['components']` or something like that :)
```
try to keep types as narrow as possible - we know upfront what does and doesn't exist, let's make the type system double check against typos etc.

src/registry/addressNL/edit.tsx
```diff
@@ -81,6 +142,15 @@ const EditForm: EditFormDefinition<AddressNLComponentSchema> = () => {
   const [isKeyManuallySetRef, generatedKey] = useDeriveComponentKey();
   const {hasAnyError} = useErrorChecker<AddressNLComponentSchema>();
   Validate.useManageValidatorsTranslations<AddressNLComponentSchema>(['required']);
+  Validate.useManageValidatorsTranslations(
+    ['pattern'],
+    `openForms.components.postcode.translatedErrors`
```
sergei-maertens, line 147: this doesn't match with the type passed in for the generic [here](https://github.com/open-formulieren/formio-builder/pull/167/files#diff-e1c3cd11dc4c171ab1f0efa43e03b841daaaab0141ce38725838bd6272ec462dR16) - and you're also not passing the sub-schema generic argument to `useManageValidatorsTranslations`

Should be something like:

```tsx
type AddressSubComponents = Required<AddressNLComponentSchema>['openForms']['components'];

Validate.useManageValidatorsTranslations<AddressSubComponents[keyof AddressSubComponents]>(...);
```

I also notice that the field name always ends with `translatedErrors`, so perhaps it's better to pass in an optional prefix for deeply nested fields instead of the entire field name? Then the generic using `S['translatedErrors']` would still enforce the same/similar shape of managing translation errors, which is nice to have.

sergei-maertens (changes requested):

src/registry/addressNL/edit.tsx
stevenbal: I'm trying to get this to work, but it's costing me quite a lot of time. I've got it "working" (no typing errors, except the build does fail) like this with these changes in types https://github.com/open-formulieren/types/pull/52/files:

```typescript
Validate.useManageValidatorsTranslations<NonNullable<AddressComponents[keyof AddressComponents]>['validate']>(
    ['pattern'], `openForms.components.postcode.translatedErrors`
  );
```

But then when building, I get:

```
src/registry/addressNL/edit.tsx:340:20 - error TS2322: Type '{ pattern: string; }' is not assignable to type 'HasValidation<"pattern">'.
  Object literal may only specify known properties, and 'pattern' does not exist in type 'HasValidation<"pattern">'.

340         validate: {pattern: ''},
```

Any ideas? :grimacing:

sergei-maertens: not immediately, I'll have to dive into this, I'll reserve some time tomorrow :grimacing:

src/components/builder/validate/i18n.tsx
```diff
@@ -9,12 +9,12 @@ import {BuilderContext} from '@/context';
 import {DataMap, Panel, Tab, TabList, TabPanel, Tabs, TextField} from '../../formio';
 
 export function useManageValidatorsTranslations<S extends SchemaWithValidation>(
-  keys: PossibleValidatorErrorKeys<S>[]
+  keys: PossibleValidatorErrorKeys<S>[],
+  field: string = 'translatedErrors'
```
sergei-maertens, line 13: so, to avoid weir things, I don't think this should be typed as `string` and even the field as a whole should be specified - probably it's better to allow a prefix (which can be anything):

```tsx
// function signature
keys: PossibelValidatorErrorKeys<S>[],
prefix: string = '',

// function body
const fieldName = `${prefix}${prefix ? '.' : ''}translatedErrors`;
```

The rest of the type information looks like it can remain as is, however [L64](https://github.com/open-formulieren/formio-builder/pull/167/files#diff-e1c3cd11dc4c171ab1f0efa43e03b841daaaab0141ce38725838bd6272ec462dR64) will need to include the prefix too I think, since the form field name must be the deeply nested sub component schema.

src/registry/addressNL/edit.stories.ts
```diff
@@ -0,0 +1,160 @@
...
+export const PostcodeValidationTabWithoutConfiguration: Story = {
+  name: 'AddressNL postcode validation tab (no prior configuration)',
+  play: async ({canvasElement, step}) => {
+    const canvas = within(canvasElement);
+
+    await step('Navigate to validation tab and open Postcode validation', async () => {
+      await userEvent.click(canvas.getByRole('link', {name: 'Validation'}));
+      await userEvent.click(canvas.getAllByText('Postcode')[0]);
+      await waitFor(async () => {
+        expect(await canvas.findByText('Regular expression for postcode')).toBeVisible();
```
sergei-maertens, line 39: do you really need to wrap all this in `waitFor`? `await canvas.findByText` is already async and internally uses `waitFor`

src/registry/addressNL/edit.stories.ts
```diff
@@ -0,0 +1,160 @@
...
+        expect(await canvas.findByText('NL')).toBeVisible();
+        expect(await canvas.findByText('EN')).toBeVisible();
+        expect(await canvas.findByText('Error code')).toBeVisible();
+        expect(await canvas.findByDisplayValue('pattern')).toBeVisible();
```
sergei-maertens, line 43: I thought I set up accessible `aria-label` attributes in these components, so you should be able to `await canvas.getByLabelText(...)` these and then assert `expect(...).toHaveDisplayValue(...)` them.

stevenbal: I'm not sure what you mean, would I be doing `canvas.getByLabelText("Error code")`? That doesn't seem to work because it doesn't have an `aria-label` (from what I can see)

src/components/formio/datamap.tsx
```diff
@@ -30,6 +32,13 @@ export const DataMap: React.FC<DataMapProps> = ({name, valueComponent, keyLabel
               const newValue = {...value, [item.key]: event.target.value};
               setValue(newValue);
             },
+            'aria-label': intl.formatMessage(
+              {
+                description: 'Accessible label for error message input field',
+                defaultMessage: 'Error message for "{field}"',
+              },
+              {field: item.key}
```
stevenbal, line 40: @sergei-maertens I'm not sure if it's possible to define the formatted message outside the `DataMap` component and then format it again with the values from the loop?

sergei-maertens: yeah you can do:

```tsx
// module level constant
const labelMsg = intl.defineMessage({
    description: '...',
    defaultMessage: 'Error message for "{field}"',
});

// ... rest of code

// in loop
intl.formatMessage(labelMsg, {field: item.key});
```

src/components/formio/datamap.tsx
```diff
@@ -8,12 +9,19 @@ export interface DataMapProps {
...
 
 /**
  * Data map with readonly keys.
  */
-export const DataMap: React.FC<DataMapProps> = ({name, valueComponent, keyLabel = 'Key'}) => {
+export const DataMap: React.FC<DataMapProps> = ({
+  name,
+  valueComponent,
+  keyLabel = 'Key',
+  ariaLabelMessage = null,
```
sergei-maertens, line 22:
```suggestion
  ariaLabelMessage,
```
I think you can just leave it as `undefined` - that's also falsy

src/components/formio/datamap.tsx
```diff
@@ -30,6 +38,9 @@ export const DataMap: React.FC<DataMapProps> = ({name, valueComponent, keyLabel
               const newValue = {...value, [item.key]: event.target.value};
               setValue(newValue);
             },
+            'aria-label': ariaLabelMessage
+              ? intl.formatMessage(ariaLabelMessage, {key: item.key})
+              : null,
```
sergei-maertens, line 43:
```suggestion
            'aria-label': ariaLabelMessage
              ? intl.formatMessage(ariaLabelMessage, {key: item.key})
              : undefined,
```
using `undefined` makes it as if the prop was not passed at all, where-as `null` treats it that the prop *was* passed.

src/registry/addressNL/edit.stories.ts
```diff
@@ -0,0 +1,152 @@
...
+
+export const PostcodeValidationTabWithoutConfiguration: Story = {
+  name: 'AddressNL postcode validation tab (no prior configuration)',
+  play: async ({canvasElement, step}) => {
+    const canvas = within(canvasElement);
+
+    await step('Navigate to validation tab and open Postcode validation', async () => {
+      await userEvent.click(canvas.getByRole('link', {name: 'Validation'}));
+      await userEvent.click(canvas.getAllByText('Postcode')[0]);
+      expect(await canvas.findByText('Regular expression for postcode')).toBeVisible();
```
sergei-maertens, line 38: I think a more complete query/assertion would be:

```suggestion
      expect(await canvas.findByLabelText('Regular expression for postcode')).toBeVisible();
```
this asserts that:

1. there is a label with this text
2. there is an input associated with the label

src/registry/addressNL/edit.stories.ts
```diff
@@ -0,0 +1,152 @@
...
+  },
+  play: async ({canvasElement, step}) => {
+    const canvas = within(canvasElement);
+
+    await step('Navigate to validation tab and open Postcode validation', async () => {
+      await userEvent.click(canvas.getByRole('link', {name: 'Validation'}));
+      await userEvent.click(canvas.getAllByText('Postcode')[0]);
+      const patternInput = canvas.getByLabelText(
+        'Regular expression for postcode'
+      ) as HTMLInputElement;
```
sergei-maertens, line 95: the `as` casting should not be needed, often you can pass the expected type as a generic param: `getByLabelText<HTMLInputElement>` *or* if the problem is that it can be `null` or `undefined`, you can add an exclamation mark at the end saying that you know it will resolve:

```tsx
canvas.getByLabelText("some label")!;
```

sergei-maertens (approved):

src/registry/addressNL/edit.stories.ts
stevenbal: I'll pass the generic param, the exclamation mark doesn't work, it gives me the same error as when I don't add the type at all: `Property 'value' does not exist on type 'HTMLElement'.ts(2339)` (on the line `expect(patternInput.value).toBe('Amsterdam');`)

PR #168, opened 2024-07-09
:speech_balloon: [open-formulieren/open-forms#4505] Change selectboxes validation placeholders

stevenbal:
partially fixes open-formulieren/open-forms#4505


sergei-maertens (approved):
I probably would have just deleted them :stuck_out_tongue: but this is better

PR #174, opened 2024-08-15
Remove the low-level fallback handling

sergei-maertens:
This is a bit overdue now, so clearing this up as I noticed I was starting to bump into it with the form preview.

The situation is:

* Open Forms only supports a subset of formio components
* Open Forms also defines some own custom components
* It's likely at some point JSON will end up in the form definition that will contain component types that we don't know/support

It's important that the application doesn't crash when such an unsupported component is present, but the challenge is that we don't know what it looks like. So, for that, `FallbackSchema` exists.

In the public API entrypoints, we must allow for `AnyComponentSchema` and `FallbackSchema`, so that our type checker helps us in preventing crashes. This was and remains the case.

This PR instead moves all the fallback handling from the internal/private API, as we are feature complete in terms of supported components and don't need this anymore. This should simplify the code/maintenance quite a bit.


src/components/ComponentConfiguration.tsx
```diff
@@ -42,34 +49,60 @@ const ComponentConfiguration: React.FC<ComponentConfigurationProps> = ({
...
+  component: FallbackSchema;
+  theme: BuilderContextType['theme'];
+}
+
+const Fallback: React.FC<FallbackProps> = ({component, theme}) => (
+  <>
+    <p>
+      The component type <code>{component.type ?? '-'}</code> is unknown. We can only show the JSON
+      definition below.
+    </p>
```
sergei-maertens, line 97: TODO: properly translate

PR #180, opened 2024-09-19
Issue/4637 optional translations act as required

robinmolen:
The translations of the option values where required because of a small issue in the zog validation configuration. This is now fixed

I've also added some tests to the componentConfigurations story, and to the stories of the individual components. Each story now has tests for the translation inputs. Ensuring that all translation inputs can be targeted, filled in, cleared and result in no errors.

### DIFF
--- a/src/components/ComponentConfiguration.stories.tsx
+++ b/src/components/ComponentConfiguration.stories.tsx
@@ -1204,6 +1204,21 @@ export const SelectBoxes: Story = {
       // check that the option labels are in the translations table
       expect(await editForm.findByText('Option label 1')).toBeVisible();
       expect(await editForm.findByText('Second option')).toBeVisible();
+
+      // Check that all translations can be filled
+      const inputs = editForm.getAllByRole('textbox');
+      for (let input of inputs) {
+        await userEvent.type(input, 'manualTranslation');
+        await expect(input).toHaveValue('manualTranslation');
+        await userEvent.clear(input);
+        await expect(input).toHaveValue('');
+      }
+
+      // Removing focus from the last input
+      await userEvent.click(canvas.getByRole('tab', {name: 'Translations'}));
+
+      // Check that none of the inputs have a Required error message
+      await expect(await editForm.queryByText('Required')).toBeNull();
     });
 
     await step('Set up itemsExpression for options', async () => {
@@ -1239,7 +1254,9 @@ export const SelectBoxes: Story = {
         openForms: {
           dataSrc: 'variable',
           itemsExpression: {var: 'someVar'},
-          translations: {},
+          translations: {
+            nl: {description: '', label: '', tooltip: ''},
+          },
         },
         defaultValue: {},
         // Advanced tab
@@ -1410,6 +1427,21 @@ export const Radio: Story = {
       // check that the option labels are in the translations table
       expect(await editForm.findByText('Option label 1')).toBeVisible();
       expect(await editForm.findByText('Second option')).toBeVisible();
+
+      // Check that all translations can be filled
+      const inputs = editForm.getAllByRole('textbox');
+      for (let input of inputs) {
+        await userEvent.type(input, 'manualTranslation');
+        await expect(input).toHaveValue('manualTranslation');
+        await userEvent.clear(input);
+        await expect(input).toHaveValue('');
+      }
+
+      // Removing focus from the last input
+      await userEvent.click(canvas.getByRole('tab', {name: 'Translations'}));
+
+      // Check that none of the inputs have a Required error message
+      await expect(await editForm.queryByText('Required')).toBeNull();
     });
 
     await step('Set up itemsExpression for options', async () => {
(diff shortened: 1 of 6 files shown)


src/components/ComponentConfiguration.stories.tsx
```diff
@@ -1204,6 +1206,21 @@ export const SelectBoxes: Story = {
       // check that the option labels are in the translations table
       expect(await editForm.findByText('Option label 1')).toBeVisible();
       expect(await editForm.findByText('Second option')).toBeVisible();
+
+      // Check that all translations can be filled
+      const inputs = editForm.getAllByRole('textbox');
+      for (let input of inputs) {
+        await userEvent.type(input, 'manualTranslation');
+        await expect(input).toHaveValue('manualTranslation');
```
sergei-maertens, line 1214: `expect` should not necessarily be `await`-ed. It shouldn't break anything, but it is a bit confusing. Typically it should be something like `expect(await canvas.findByRole(...)).toBeVisible()` to account for some rendering delays etc.

these existing tests are littered with bad examples, so I can hardly blame you for this.

src/components/ComponentConfiguration.stories.tsx
```diff
@@ -1204,6 +1206,21 @@ export const SelectBoxes: Story = {
...
+        await userEvent.clear(input);
+        await expect(input).toHaveValue('');
```
sergei-maertens, line 1216:
```suggestion
        expect(input).toHaveValue('');
```

src/components/ComponentConfiguration.stories.tsx
sergei-maertens, line 1214:
```suggestion
        expect(input).toHaveValue('manualTranslation');
```

src/components/ComponentConfiguration.stories.tsx
```diff
@@ -1204,6 +1206,21 @@ export const SelectBoxes: Story = {
...
+      }
+
+      // Removing focus from the last input
+      await userEvent.click(canvas.getByRole('tab', {name: 'Translations'}));
+
+      // Check that none of the inputs have a Required error message
+      await expect(await editForm.queryByText('Required')).toBeNull();
```
sergei-maertens, line 1223:
```suggestion
      expect(await editForm.queryByText('Required')).toBeNull();
```

src/registry/addressNL/edit.stories.ts
```diff
@@ -35,10 +35,14 @@ export const PostcodeValidationTabWithoutConfiguration: Story = {
     await step('Navigate to validation tab and open Postcode validation', async () => {
       await userEvent.click(canvas.getByRole('link', {name: 'Validation'}));
       await userEvent.click(canvas.getAllByText('Postcode')[0]);
-      expect(await canvas.findByLabelText('Regular expression for postcode')).toBeVisible();
-      expect(await canvas.findByText('NL')).toBeVisible();
-      expect(await canvas.findByText('EN')).toBeVisible();
-      expect(await canvas.findByText('Error code')).toBeVisible();
+
+      await waitFor(async () => {
```
sergei-maertens, line 39: it's a bit strange that here the `waitFor` is needed, since the `await canvas.findByLabelText(...)` is already async and uses `waitFor` under the hood :thinking: 

I'd like to better understand if/what is going wrong in these interaction tests - often an actual root cause can be found.

src/registry/radio/radio-validation.stories.ts
```diff
@@ -52,8 +52,42 @@ export const ManualMinimumOneValue: Story = {
...
+
+      await waitFor(async () => {
+        await expect(
+          await canvas.findByText('The option label is a required field.')
+        ).toBeVisible();
+        await expect(
+          await canvas.findByText('The option value is a required field.')
+        ).toBeVisible();
+      });
+    });
```
sergei-maertens, line 64: these often fail locally in storybook when you don't have focus on the browser window. could you check if they're really necessary? I can't really remember these tests failing in CI, while the file size validation tests are notable sources of flakiness

sergei-maertens (commented):
Nice to see extensive testing being applied!

My general theme of feedback is about the existing tests and sync/async behaviours, which unfortunately is quite tied to exactly how one holds Storybook...

Another thing that caught my attention was the amount of commit messages - once the PR is ready for check-in, could you rebase the commits to be a bit more self-contained? E.g. commits for prettier can easily be squashed in their respective commits that do the bugfix/add the (regression) test/implement a feature.

In Open Forms we rely on the commit history _really hard_ to debug and understand code, so it's important that we see "the useful" information in a git blame.

src/registry/radio/radio-validation.stories.ts
rvdm2101: Its a bit weird/scary that some tests (only) fail locally.. I'll check them out and remove the `await`'s and `waitFor`'s that aren't necessary

src/registry/addressNL/edit.stories.ts
rvdm2101: Without the `waitFor` they keep failing locally..

Maybe this is again one of those situations that only fail locally, and that the CI will work fine..

sergei-maertens (approved):

PR #181, opened 2024-09-20
add product price component

Floris272:
Adds the product price component.

this component will be able to select a product which will be used to get its current prices when a user opens the form.

"getProducts" is added to the BuilderContext to be able to fetch the products for the product select.

### DIFF
--- a/src/components/ComponentConfiguration.stories.tsx
+++ b/src/components/ComponentConfiguration.stories.tsx
@@ -9,6 +9,7 @@ import {
   DEFAULT_COLORS,
   DEFAULT_DOCUMENT_TYPES,
   DEFAULT_FILE_TYPES,
+  DEFAULT_PRODUCTS,
 } from '@/tests/sharedUtils';
 import {AnyComponentSchema} from '@/types';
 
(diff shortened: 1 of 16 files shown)


src/components/ComponentConfiguration.stories.tsx
```diff
@@ -2226,6 +2228,76 @@ export const NpFamilyMembers: Story = {
...
+export const ProductPrice: Story = {
+  render: Template,
+  name: 'type: productPrice',
+
+  args: {
+    component: {
+      id: 'wqimsadk',
+      type: 'productPrice',
+      key: 'productPrice',
+      label: 'An productPrice field',
```
sergei-maertens, line 2240:
```suggestion
      label: 'A productPrice field',
```

src/components/ComponentConfiguration.stories.tsx
```diff
@@ -2226,6 +2228,76 @@ export const NpFamilyMembers: Story = {
...
+      weight: 10,
+      schema: {},
+    },
+  },
+
+  play: async ({canvasElement, step, args}) => {
+    const canvas = within(canvasElement);
+    const preview = within(canvas.getByTestId('componentPreview'));
+
+    await expect(canvas.getByLabelText('Label')).toHaveValue('An productPrice field');
```
sergei-maertens, line 2257:
```suggestion
    await expect(canvas.getByLabelText('Label')).toHaveValue('A productPrice field');
```

src/registry/productPrice/edit.tsx
```diff
@@ -0,0 +1,165 @@
...
+  label: '',
+  key: '',
+  description: '',
+  tooltip: '',
+  showInSummary: true,
+  showInEmail: false,
+  showInPDF: true,
+  hidden: false,
+  clearOnHide: true,
+  isSensitiveData: true,
```
sergei-maertens, line 118:
```suggestion
  isSensitiveData: false,
```
it's not personally identifiable data, so I think we can go with `false`

sergei-maertens (approved):
Very nice!

PR #183, opened 2024-09-25
:bug: [open-formulieren/open-forms#4636] Preventing large values from breaking the table

robinmolen:
Closes open-formulieren/open-forms#4636

All data a user can input shouldn't be allowed to break the table

### DIFF
--- a/src/components/builder/i18n.tsx
+++ b/src/components/builder/i18n.tsx
@@ -57,7 +57,10 @@ export function ComponentTranslations<S extends AnyComponentSchema>({
       <table className="table table-bordered offb-table">
         <thead>
           <tr className="offb-i18n-header">
-            <td colSpan={2} className="offb-i18n-header__label">
+            <td
+              colSpan={2}
+              className="offb-i18n-header__label offb-table__col offb-table__col--width-50"
+            >
               <ComponentLabel
                 label={
                   <FormattedMessage
@@ -71,7 +74,7 @@ export function ComponentTranslations<S extends AnyComponentSchema>({
                 })}
               />
             </td>
-            <td className="offb-i18n-header__tab-container">
+            <td className="offb-i18n-header__tab-container offb-table__col offb-table__col--width-50">
               <ul className={`nav nav-tabs offb-i18n-header__tabs`}>
                 {supportedLanguageCodes.map(code => (
                   <li key={code} className={clsx('nav-item', {active: code === activeLanguage})}>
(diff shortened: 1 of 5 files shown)


src/components/builder/i18n.tsx
```diff
@@ -122,7 +122,10 @@ export function ComponentTranslations<S extends AnyComponentSchema>({
                 </span>
               </td>
               <td>
-                <div aria-describedby={`component-translation-property-${property}`}>
+                <div
+                  aria-describedby={`component-translation-property-${property}`}
+                  style={{wordBreak: 'break-word'}}
```
sergei-maertens, line 127: please use proper CSS for this instead of inline styles

`<div className="offb-content offb-content--allow-break">` seems a proper BEM approach:

* block 'content' for "plain" text content, so you can use this in multiple places
* modifier `allow-break` to explicitly signify we expect longs words to break

However, keep in mind that `word-break: break-word` is deprecated (see https://developer.mozilla.org/en-US/docs/Web/CSS/word-break) - instead you should probably use `overflow-wrap: break-word;`, see https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap

sergei-maertens (changes requested):

src/components/builder/i18n.tsx
robinmolen: I've named this `offb-table` and `offb-table__content--allow-break`, because this also needed some styling for the table element.

And, because `word-break: break-word` is deprecated, i've changed it to `overflow-wrap: break-word`

src/components/builder/i18n.tsx
```diff
@@ -122,7 +123,10 @@ export function ComponentTranslations<S extends AnyComponentSchema>({
                 </span>
               </td>
               <td>
-                <div aria-describedby={`component-translation-property-${property}`}>
+                <div
+                  aria-describedby={`component-translation-property-${property}`}
+                  className="offb-table__content--allow-break"
```
sergei-maertens, line 128:
```suggestion
                  className="offb-table__content offb-table__content--allow-break"
```
proper BEM naming requires the block/element to always be specified, and then you can add zero or more modifiers

I hope I can add the BEM SASS utilities here at some point too since they enforce the proper classnames in the DOM :)

src/components/builder/i18n.tsx
```diff
@@ -12,6 +12,7 @@ import {AnyComponentSchema} from '@/types/schemas';
 import {Component, TextField} from '../formio';
 import ComponentLabel from '../formio/component-label';
 import './i18n.scss';
+import './table.scss';
```
sergei-maertens, line 15: since this isn't really tied to a react component, maybe we should perform this import once in the `index.ts` of the library? I'm not sure what the right approach is here

Possibly we can also use `@use './table';` in the `i18n.scss` module to ensure it's loaded :thinking:

robinmolen: Mhh in the `index.ts` might be better yeah.. That we treat this as global styling. Because, indeed, it currently isn't tied to any component.

sergei-maertens: reverted because storybook itself doesn't hit the `index.ts` and then the styles are absent in the chromatic snapshots

sergei-maertens (approved):

PR #184, opened 2024-09-26
:bug: [open-formulieren/open-forms#4659] Ensure that textfield have a valid empty value

### LINKED ISSUE #4659: Hidden components with "Multiple values" trigger validation
### Product versie / Product version

latest

### Customer reference

no particular but it get's reported 

### Omschrijf het probleem / Describe the bug

Hidden component with "Multiple values" set get their defaultValue changed from:
```json 
  "defaultValue": null,
```
to:  

```json
  "defaultValue": [
    null
  ],
```

This causes validation to trigger right after loading since their value is now [] instead of [ null ]

```josn 
  "telefoonnummer": []
```

This can be resolved by changing the defaultValue to via the json of the components:

```json
  "defaultValue": [],
```

robinmolen:


robinmolen:
If we only want to fix the null value when the `multiple values` option is triggered, we should apply the fix to `src/components/builder/multiple.tsx`.

sergei-maertens:
Discussed that this doesn't fix the source of the problem, as it works fine in isolation. A different [approach](https://github.com/open-formulieren/open-forms/issues/4659#issuecomment-2389031875) will be done.

PR #188, opened 2024-10-14
Add support for soft required validation in builder

sergei-maertens:
Partially closes open-formulieren/open-forms#4546 - depends on https://github.com/open-formulieren/types/pull/55

* Refactored RichText component from `content` to generic builder components
* Organized ordering of storybook stories/"folders"
* Added support for soft required validation to file component
* Added new component `SoftRequiredErrors` that will collect all errors and display them. It supports rich text formatting.


src/components/builder/validate/soft-required.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const {values, setFieldValue} = useFormikContext<ComponentWithRequiredOptions>();
+
+  const isRequired = values?.validate?.required ?? false;
+  const isSoftRequired = values?.openForms?.softRequired ?? false;
+
+  // if the field is hard required, we must disable the soft required option, and
+  // synchronize the softRequired to uncheck the option (if needed)
+  useEffect(() => {
+    if (!(isRequired && isSoftRequired)) return;
+    setFieldValue('openForms.softRequired', false);
```
robinmolen, line 27: It's a bit complicated... Maybe this will be simpler:
```suggestion
    if (isRequired && isSoftRequired) {
      setFieldValue('openForms.softRequired', false);
    }
```

src/registry/softRequiredErrors/edit.tsx
```diff
@@ -0,0 +1,122 @@
...
+  useEffect(() => {
+    const currentValue = values.openForms?.translations?.[defaultLanguageCode]?.html;
+    if (currentValue === undefined && values.html) {
+      // if we have a 'html' value, but no default translation -> store the default translation
+      setFieldValue(`openForms.translations.${defaultLanguageCode}.html`, values.html);
+    } else if (values.html !== currentValue) {
+      // otherwise sync the value of the translation field to the main field.
+      setFieldValue('html', currentValue);
+    }
+  });
```
robinmolen, line 40: I'm not entirely sure what the consequences are, but shouldn't we define the dependencies to make sure this useEffect is only triggered when the `values` or the `setFieldValue` change?

src/components/builder/validate/soft-required.tsx
sergei-maertens: yeah I'm a big fan of guard clauses myself, but this reads better :)

src/registry/softRequiredErrors/edit.tsx
sergei-maertens: this function just executes on every render (it's copied from the `content` component), so at worst the hook runs a couple times too often.

Normally I'd indeed optimize this by specifying dependencies, but `values` has a stable identity (I think) from Formik, and if it didn't, then the hook would run every time anyway. We have the proper conditions that check if an update is needed or not, so you don't get infinite loops :)

robinmolen (approved):
Looks good to me :)

PR #193, opened 2024-11-08
Added prefill tab to number component edit

robinmolen:
Partly solves open-formulieren/open-forms#4813

Allow prefill configuration on number fields


vaszig (approved):
Seems ok to me!I guess we do not need any stories for this so I believe it's ready to go when we update the PR with the new version of the types repository. Let's check one more time the CI just to be sure.

robinmolen:
Yeah this is set to draft due to the types repository.

robinmolen:
I've updated the types dependency and added prefill config to the `number type` storybook test.

If you are still ok with this, then we can merge it :)

sergei-maertens (approved):

sergei-maertens:
Merged, but we will do a release later this week since I'd like to have #194 and #195 in the release too, and some pending cosign improvements work (#192)

PR #195, opened 2024-11-12
Handle API serializer renames in the backend

sergei-maertens:
Part of open-formulieren/open-forms#4606 - needs to be part of open-formulieren/open-forms#4817

The backend makes the API serializers more uniformly English in 3.0


sergei-maertens:
Merging as https://github.com/open-formulieren/open-forms/pull/4817 got approved

PR #197, opened 2024-12-03
Add background image configuration to map component

robinmolen:
Part of open-formulieren/open-forms#2173

The map component now has background configuration. This changes the image which is used for the leaflet tile layer (the background of the map component)


robinmolen:
This depends on a change in the type library. Keeping this as draft, so these changes can be added

vaszig (approved):
I don't see something weird. I didn't manage though to test this manually (tried with `npm pack` and then installed it in the backend) so I guess @sergei-maertens has to do another review here.

src/context.ts
```diff
@@ -48,6 +59,7 @@ export interface BuilderContextType {
   getDocumentTypes: () => Promise<Array<DocumentTypeOption>>;
   getConfidentialityLevels: () => Promise<SelectOption[]>;
   getAuthPlugins: () => Promise<AuthPluginOption[]>;
+  leafletMapBackgrounds: LeafletMapOption[];
```
sergei-maertens, line 62: I'd stick to using a callback for consistency with the other callbacks like `getFileTypes`, and also leave 'leaflet' out of the name. If we switch map provider, the tile layer URL will stay the same and then the property names are just confusing.

```suggestion
  getMapTileLayers: () => Promise<MapTileLayerOption[]>;
```

src/registry/map/edit.tsx
```diff
@@ -174,4 +178,31 @@ const UseConfigDefaultMapSettings: React.FC<{}> = () => {
   );
 };
 
+const Background: React.FC = () => {
```
sergei-maertens, line 181:
```suggestion
const TileLayer: React.FC = () => {
```

sergei-maertens (changes requested):

src/context.ts
robinmolen: Okay, yeah sounds good.

Yeah the leaflet part also doesn't mean a whole lot here.. I'll change the naming in all repositories to reflect this

src/context.ts
```diff
@@ -48,6 +59,7 @@ export interface BuilderContextType {
   getDocumentTypes: () => Promise<Array<DocumentTypeOption>>;
   getConfidentialityLevels: () => Promise<SelectOption[]>;
   getAuthPlugins: () => Promise<AuthPluginOption[]>;
+  getMapTileLayers: () => Promise<MapTileLayers[]>;
```
sergei-maertens, line 62:
```suggestion
  getMapTileLayers: () => Promise<MapTileLayer[]>;
```

src/registry/map/edit.tsx
```diff
@@ -174,4 +179,36 @@ const UseConfigDefaultMapSettings: React.FC<{}> = () => {
...
+  const {getMapTileLayers} = useContext(BuilderContext);
+  const {value: options, loading, error} = useAsync(async () => await getMapTileLayers(), []);
+  if (error) {
+    throw error;
+  }
+
+  const intl = useIntl();
+  const tooltip = intl.formatMessage({
+    description: "Tooltip for 'tileLayerIdentifier' builder field",
+    defaultMessage: 'The tile layer to use as the map.',
```
sergei-maertens, line 192:
```suggestion
    defaultMessage: 'The tile layer is responsible for showing the map parts at particular coordinates and zoom levels.',
```

src/registry/map/preview.tsx
```diff
@@ -37,10 +39,27 @@ const Preview: React.FC<ComponentPreviewProps<MapComponentSchema>> = ({component
...
+  }
+
   const {required = false} = validate;
   const {lat = 52.1326332, lng = 5.291266} = initialCenter;
   const zoom = defaultZoom ?? 8;
+
+  const tileLayerUrl = (): string => {
+    if (!tileLayerIdentifier || loading || !options?.length) {
+      return TILE_LAYER_RD.url;
+    }
```
sergei-maertens, line 57: I think it's a bit odd the tile layer can change because we're still loading them. Perhaps it's nicer to just display a loader rather than any map at all while we're still fetching the available tile layers? (referring to the `loading` var)

src/registry/map/preview.tsx
```diff
@@ -37,10 +39,27 @@ const Preview: React.FC<ComponentPreviewProps<MapComponentSchema>> = ({component
...
+    return (
+      options?.find(tileLayer => tileLayer.identifier === tileLayerIdentifier)?.url ??
+      TILE_LAYER_RD.url
+    );
```
sergei-maertens, line 61: this should cover `options` being an empty array too, right? And even `tileLayerIdentifier` being `undefined`

src/tests/sharedUtils.tsx
```diff
@@ -1,5 +1,5 @@
 // This module contains shared utilities and constants between Jest and Storybook.
-import type {DocumentTypeOption, SelectOption} from '@/context';
+import type {DocumentTypeOption, MapTileLayers, SelectOption} from '@/context';
```
sergei-maertens, line 2:
```suggestion
import type {DocumentTypeOption, MapTileLayer, SelectOption} from '@/context';
```

src/tests/sharedUtils.tsx
```diff
@@ -235,6 +235,19 @@ export const DEFAULT_COLORS: ColorOption[] = [
   },
 ];
 
+export const DEFAULT_MAP_TILE_LAYERS: MapTileLayers[] = [
```
sergei-maertens, line 238:
```suggestion
export const DEFAULT_MAP_TILE_LAYERS: MapTileLayer[] = [
```

src/context.ts
```diff
@@ -29,6 +29,17 @@ export interface DocumentTypeOption {
   description: string;
 }
 
+/*
+  Map tile layers
+
+  This datastructure is created by the Open Forms backend.
+ */
+export interface MapTileLayers {
```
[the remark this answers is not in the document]
robinmolen: Ah shit yeah... my bad..

src/registry/map/edit.tsx
robinmolen: Mhh, yeah i agree that the defaultMessage is pretty empty right now.. (and should be changed).
But i'm not sure if we should mention the coordinates and zoom levels. It isn't really connected to each other..

Maybe something like `The tile layer determines the used map background.` is more fitting

src/registry/map/preview.tsx
robinmolen: Those checks are done on line 55. With the `!tileLayerIdentifier` for undefined (or otherwise falsey values), and `!options?.length` for undefined or empty (array size 0). If either one of those is true the function will return the `TILE_LAYER_RD.url` prematurely

src/registry/map/preview.tsx
robinmolen: Check, yeah good point.

I've added a simple Loader component, using the basic bootstrap spinner. This will be shown, when the data is still loading.

src/context.ts
robinmolen: I've fixed this

src/tests/sharedUtils.tsx
robinmolen: check, its been updated/fixed

src/components/builder/loader.stories.tsx
```diff
@@ -0,0 +1,23 @@
...
+    controls: {hideNoControlsWarning: true},
+    docs: {
+      source: {
+        type: 'dynamic',
+        excludeDecorators: true,
+      },
+    },
+    modal: {noModal: true},
+  },
+} as Meta<typeof Loader>;
```
sergei-maertens, line 18:
```suggestion
} satisfies Meta<typeof Loader>;
```
this should be preferred since it doesn't just cast the type, it performs the proper type checks still :)

src/components/builder/loader.stories.tsx
```diff
@@ -0,0 +1,23 @@
...
+  title: 'Formio/Builder/Loader',
+  component: Loader,
+  parameters: {
+    controls: {hideNoControlsWarning: true},
+    docs: {
+      source: {
+        type: 'dynamic',
+        excludeDecorators: true,
+      },
+    },
```
sergei-maertens, line 15: I don't think this is needed anymore, it was a Storybook 6/7 thing when using MDX for documentation.

src/components/builder/loader.stories.tsx
```diff
@@ -0,0 +1,23 @@
...
+    modal: {noModal: true},
+  },
+} as Meta<typeof Loader>;
+
+export const Default: StoryObj<typeof Loader> = {
+  render: () => <Loader />,
+  args: {},
+};
```
sergei-maertens, line 23:
```suggestion
type Story = StoryObj<typeof Loader>;

export const Default: Story = {};
```
defining the type alias outside makes it a bit more re-usable/DRY and the `render` override isn't needed since that's what SB does by default.

src/registry/map/edit.tsx
```diff
@@ -174,4 +179,38 @@ const UseConfigDefaultMapSettings: React.FC<{}> = () => {
...
 };
 
+const TileLayer: React.FC = () => {
+  const {getMapTileLayers} = useContext(BuilderContext);
+  const {value: options, loading, error} = useAsync(async () => await getMapTileLayers(), []);
+  if (error) {
+    throw error;
+  }
+
+  const intl = useIntl();
```
sergei-maertens, line 189: you must call `useIntl()` before possibly throwing an error, since hooks cannot be called conditionally

```suggestion
  const intl = useIntl();

  if (error) {
    throw error;
  }
```

src/registry/map/preview.tsx
robinmolen: Ah check, yeah i've changed this a bit, by removing the `?.` from line 58 (because, indeed, at that point options is at least defined as an array).

Yeah your assumption is right. The guard clause isn't really needed..
I do prefer to check the validity of `tileLayerIdentifier` outside the `.find()`. If we don't have to loop over data, we probably shouldn't. I also kinda like a early return, but in this case it might be a bit extreme..

So yeah, we can remove the guard and turn this into a one-liner :)

src/registry/map/edit.tsx
robinmolen: Ah shit, you're right..

src/components/builder/loader.stories.tsx
robinmolen: Check, I've updated it

src/components/builder/loader.stories.tsx
robinmolen: Ah okay cool! Yeah that sounds a lot better 👌

src/components/builder/loader.stories.tsx
robinmolen: Check, it indeed wasn't needed anymore.. Ive removed it

src/registry/map/preview.tsx
robinmolen: Ive turned it into a one-liner

sergei-maertens (approved):
Good to merge, and then bag it and tag it! :rocket:

PR #198, opened 2024-12-11
Adding map interaction config

robinmolen:
Part of open-formulieren/open-forms#2177

This relies on a update to the types repo: https://github.com/open-formulieren/types/pull/62


robinmolen:
Because this feature will be added after 3.0, this PR is moved to draft.

This PR is still ready to be reviewed, it just shouldn't be merged just yet

vaszig (commented):
As far as I can say, it seems good to me. I would add though some stories for the interactions (not necessarily tests, but just to have the appearance).

robinmolen:
> As far as I can say, it seems good to me. I would add though some stories for the interactions (not necessarily tests, but just to have the appearance).

The map component stories do show the configuration, but the choice doesn't have a visual representation.
But yeah, it would probably be nice to show/test the interaction options in the builder..

robinmolen:
This PR can be reviewed (this includes some chromatic changes that also have to be reviewed)

In the formio-builder a new map component configuration option was added: `interactions`. "Interactions" is a group of checkboxes, allowing different user interactions in the map. To keep the logic close to what we already have, the `pointer` interaction is activated by default.

In the form builder map component preview you can test/use the active interactions.

i18n/messages/nl.json
```diff
@@ -151,6 +151,11 @@
     "isTranslated": true,
     "originalDefault": "CSS class"
   },
+  "3xnsUT": {
+    "defaultMessage": "Lijn interacties",
```
sergei-maertens, line 155:
```suggestion
    "defaultMessage": "Lijn",
```
I feel like "interactions" is repeated way too much. I'm also inclined to rephrase this as "available drawing shapes" to make it simpler?

i18n/messages/nl.json
```diff
@@ -572,6 +577,11 @@
     "description": "Label for 'disableAddingRemovingRows' builder field",
     "originalDefault": "Disable adding or removing groups"
   },
+  "NbDr3m": {
+    "defaultMessage": "Geeft gebruikers de mogelijkheid een lijn te tekenen, wanneer ze met het kaartmateriaal werken",
```
sergei-maertens, line 581:
```suggestion
    "defaultMessage": "Gebruikers kunnen rechte lijnen tekenen op de kaart.",
```

sergei-maertens: is it one or more lines, or just one? Probably good to clarify that in the tooltip.

i18n/messages/nl.json
```diff
@@ -598,6 +608,11 @@
     "isTranslated": true,
     "originalDefault": "Label"
   },
+  "POkaFQ": {
+    "defaultMessage": "Polygoon interacties",
```
sergei-maertens, line 612:
```suggestion
    "defaultMessage": "Veelhoek (polygoon)",
```

i18n/messages/nl.json
```diff
@@ -885,6 +900,11 @@
     "description": "Tooltip for 'includePartners' builder field",
     "originalDefault": "Whether to add partners information to the component."
   },
+  "ZRM6J8": {
+    "defaultMessage": "Mogelijke kaartmateriaal interacties",
```
sergei-maertens, line 904:
```suggestion
    "defaultMessage": "Mogelijke kaartmateriaalinteracties",
```
single samengesteld woord

i18n/messages/nl.json
```diff
@@ -1158,6 +1178,11 @@
     "description": "Label for 'ClearOnHide' builder field",
     "originalDefault": "Clear on hide"
   },
+  "iyWQAt": {
+    "defaultMessage": "Geeft gebruikers de mogelijkheid een polygoon te tekenen, wanneer ze met het kaartmateriaal werken",
```
sergei-maertens, line 1182:
```suggestion
    "defaultMessage": "Gebruikers kunnen een veelhoek (gesloten vorm die uit rechte lijnen bestaat) tekenen.",
```

i18n/messages/nl.json
```diff
@@ -1199,6 +1224,11 @@
     "description": "Label for translation message for validation error code",
     "originalDefault": "Error message"
   },
+  "kIX7bm": {
+    "defaultMessage": "Pin interacties",
```
sergei-maertens, line 1228:
```suggestion
    "defaultMessage": "Pin/punt",
```

i18n/messages/nl.json
```diff
@@ -1464,6 +1494,11 @@
     "description": "Tooltip for 'hideHeader' builder field",
     "originalDefault": "Do not display the configured label and top line as the header in the fieldset."
   },
+  "vFydxX": {
+    "defaultMessage": "Geeft gebruikers de mogelijkheid een pin te plaatsen, wanneer ze met het kaartmateriaal werken",
```
sergei-maertens, line 1498:
```suggestion
    "defaultMessage": "Gebruikers kunnen een punt aanduiden op de kaart.",
```

src/jsonEditor.scss
```diff
@@ -1,3 +1,6 @@
+@import 'leaflet-draw/dist/leaflet.draw.css';
+@import 'leaflet/dist/leaflet.css';
```
sergei-maertens, line 2: why are these imports in an unrelated scss file? Can't you just import them close to where they're used (https://github.com/open-formulieren/formio-builder/blob/main/src/registry/map/index.ts)? And preferably a `preview.scss` file that loads these libraries, so we have a consistent pattern of loading stylesheets for a given context: https://github.com/open-formulieren/formio-builder/blob/2d64b61d74ef79145a401dbe22debec0adef36cc/src/registry/email/preview.tsx#L7

src/registry/map/preview.tsx
```diff
@@ -84,6 +93,24 @@ const Preview: React.FC<ComponentPreviewProps<MapComponentSchema>> = ({component
...
       >
         <TileLayer {...TILE_LAYER_RD} url={tileLayerUrl()} />
+        <FeatureGroup ref={featureGroupRef}>
+          <EditControl
+            position="topright"
+            onCreated={onFeatureCreate}
+            edit={{
+              edit: false,
+              remove: false,
+            }}
```
sergei-maertens, line 103: what do these control? Does it mean once I've drawn something, I can't remove it anymore from the map?

tsconfig.json
```diff
@@ -18,6 +18,7 @@
     "strictNullChecks": true,
     "allowSyntheticDefaultImports": true,
     "noErrorTruncation": true,
+    "skipLibCheck": true,
```
sergei-maertens, line 21: what's the impact of this? I don't like this part from the documentation:

> ...at the expense of type-system accuracy

sergei-maertens (changes requested):
Can we move *all* of the map settings to a separate tab? Tab title can be "Map settings" or just "Settings", and would come after the "Validation" tab. I think the "Basic" tab is getting too clutter.

Specifically, this means the fields:

* Use globally configured map component settings
* Map configuration (and re-label this to "Initial focus" or something like that)
* Tile layer
* Available interactions

src/registry/map/preview.tsx
robinmolen: Yeah they add delete and edit buttons to the map. They are a bit cumbersome when using, so i opted to remove them..
At the moment only 1 shape is allowed. So when you create a new one, the old shape is removed. This does mean that once you have added a shape, you cannot clear it..

Im taking a look if we can change the behavior of these buttons (i.e. remove all shapes when you click the button, instead of having to click to shape and then save the deletion). For reference, this is how it normally looks https://leaflet.github.io/Leaflet.draw/docs/examples/full.html

sergei-maertens: Yeah not being able to make corrections is a problem. I think it's probably best to enable the delete, but leave the edit disabled and I'm sure there will be accessibility feedback about this in the future.

tsconfig.json
robinmolen: This skips type checking of *.d.ts files. With this, only the *.d.ts files that we directly use are type checked.

This is a 'solution' for a typescript error originating from `node_modules/react-leaflet-draw/src/index.d.ts`.

As we discussed in the office, maybe better to isolate this, or to solve this another way. I've gone trough the repo/issues of react-leaflet-draw, but i haven't found anything related to this.. So i created a ticket, asking their advice about using typescript https://github.com/alex3165/react-leaflet-draw/issues/188

robinmolen: I've created a PR for this issue: https://github.com/alex3165/react-leaflet-draw/pull/189. Hopefully this gets resolved soon, so that we can remove the `"skipLibCheck": true`.

I have added a issue to the backlog to clean this up https://github.com/open-formulieren/open-forms/issues/5017

i18n/messages/nl.json
robinmolen: Yeah 'shapes' is probably better/easier to describe this that 'interactions'..

i18n/messages/nl.json
```diff
@@ -804,6 +824,11 @@
     "description": "Tooltip for 'Multiple values' builder field",
     "originalDefault": "Are there multiple values possible for this field?"
   },
+  "WXpj4j": {
+    "defaultMessage": "Gebruikers kunnen een punt aanduiden op de kaart",
```
sergei-maertens, line 828:
```suggestion
    "defaultMessage": "Gebruikers kunnen een punt aanduiden op de kaart.",
```
ironic that the "punt" sentence is missing a "punt" :laughing:

i18n/messages/nl.json
```diff
@@ -1484,6 +1519,11 @@
     "description": "Tooltip for 'prefill.attribute' builder field",
     "originalDefault": "Specify the attribute holding the pre-fill data."
   },
+  "wJ7Yha": {
+    "defaultMessage": "Kaartmateriaal instellingen",
```
sergei-maertens, line 1523:
```suggestion
    "defaultMessage": "Kaartinstellingen",
```

i18n/messages/nl.json
```diff
@@ -572,6 +582,11 @@
     "description": "Label for 'disableAddingRemovingRows' builder field",
     "originalDefault": "Disable adding or removing groups"
   },
+  "N2dWW9": {
+    "defaultMessage": "Mogelijke kaartmateriaalvormen",
```
sergei-maertens, line 586:
```suggestion
    "defaultMessage": "Beschikbare vormen",
```
maybe?

src/registry/map/edit.tsx
```diff
@@ -139,6 +155,11 @@ EditForm.defaultValues = {
     lng: undefined,
   },
   tileLayerIdentifier: undefined,
+  interactions: {
+    polygon: false,
+    polyline: false,
+    marker: true,
+  },
```
sergei-maertens, line 162: This will require a data migration and import patcher in the backend - do we have that?

src/registry/map/interaction-configuration.tsx
```diff
@@ -0,0 +1,80 @@
...
+      tooltip={tooltip}
+    />
+  );
+};
+
+const InteractionConfiguration: React.FC = () => (
+  <Panel
+    title={
+      <FormattedMessage
+        description="Interaction configuration panel title"
```
sergei-maertens, line 69:
```suggestion
        description="Map interaction configuration panel title"
```
keep in mind that messages are global so you need to provide sufficient context in the description for translators :)

src/registry/map/map-configuration.tsx
```diff
@@ -85,7 +85,7 @@ const MapConfiguration: React.FC = () => (
     title={
       <FormattedMessage
         description="Map configuration panel title"
-        defaultMessage="Map configuration"
+        defaultMessage="Initial focus"
```
sergei-maertens, line 88: Probably best to also update the description to make it a bit more accurate. Translation files need to be updated anyway.

tsconfig.json
```diff
@@ -18,6 +18,11 @@
     "strictNullChecks": true,
     "allowSyntheticDefaultImports": true,
     "noErrorTruncation": true,
+    // The `skipLibCheck` is needed to hide a typescript issue in react-leaflet-draw
+    // see issue for context https://github.com/alex3165/react-leaflet-draw/issues/188.
+    // When the typescript issue in react-leaflet-draw,
+    // we should update that dependency and remove this config.
+    "skipLibCheck": true,
```
sergei-maertens, line 25: let's include the patch-package patch in this PR as well, following our Slack discussion.

src/registry/map/interaction-configuration.tsx
robinmolen: Ah yeah, check

tsconfig.json
robinmolen: Yes, i've added it 🙂

i18n/messages/nl.json
robinmolen: Yeah, sounds good

src/registry/map/map-configuration.tsx
robinmolen: Check, yeah probably best.. Will do

src/registry/map/edit.tsx
robinmolen: The import is a good one, that is indeed needed..

robinmolen: Like we discussed via slack, i've added migration converters and a database migration script. The migration converters get used by the import, ensuring map components have the interactions configuration

PR #204, opened 2025-01-14
:sparkles: [#5016] Referentielijsten dataSrc for options

stevenbal:
closes open-formulieren/open-forms#5016 partially

~~Requires https://github.com/open-formulieren/types/pull/64 to be merged and released~~


src/components/builder/values/values-config.stories.tsx
```diff
@@ -300,3 +303,57 @@ export const SelectVariable: SelectStory = {
     },
   },
 };
+
+export const SelectReferentielijsten: SelectStory = {
+  ...Select,
+
+  decorators: [withFormik, BuilderContextDecorator],
```
sergei-maertens, line 310: I think we can move this up into the meta instead, I'm not sure why I ever decided this should go in a base story thing :thinking: 

They all need access to formik afaik.

src/components/formio/select.tsx
```diff
@@ -35,6 +35,12 @@ export interface SelectProps<
   onChange?: (event: {target: {name: string; value: ValueType}}) => void;
 }
 
+export interface Option {
+  value: string;
+  label: React.ReactNode;
+  description?: string;
+}
```
sergei-maertens, line 42: See the remark on line 20 :grimacing: 

The shape of react select options doesn't really matter, anything goes as long as you can point react-select to which property holds the actual 'value' and provide a callback/component to render the label given an option. By default, it looks at `value` and `label`, but it's actually super flexible.

Certain fields are just different than others and have different types of options that are relevant for them, and that's okay.

src/registry/radio/edit-validation.ts
```diff
@@ -14,7 +14,11 @@ const buildValuesSchema = (intl: IntlShape) =>
...
     values: optionSchema(intl).array().min(1).optional(),
     openForms: z.object({
-      dataSrc: z.union([z.literal('manual'), z.literal('variable')]),
+      dataSrc: z.union([
+        z.literal('manual'),
+        z.literal('variable'),
+        z.literal('referentielijsten'),
+      ]),
       // TODO: wire up infernologic type checking
       itemsExpression: jsonSchema.optional(),
```
sergei-maertens, line 23: this is missing the `code` and `service` (optional) schema's, no?

src/registry/radio/helpers.ts
```diff
@@ -9,3 +10,27 @@ export const checkIsManualOptions = (
...
   return component.openForms.dataSrc === 'manual';
 };
+
+// A type guard is needed because TS cannot figure out it's a discriminated union
+// when the discriminator is nested.
+// See https://github.com/microsoft/TypeScript/issues/18758
+export const checkIsReferentielijstenOptions = (
+  component: RadioComponentSchema
+): component is RadioComponentSchema & {
+  data: {values: Option[] | undefined};
```
sergei-maertens, line 20: for the builder, this means that the `values` key will be empty or missing in the radio component schema. For the renderer, it will be populated, but this is the builder here :) Also, the `values` key is directly included in radio, only `select` nests this under the `data` intermediate key.

```suggestion
```

src/registry/radio/helpers.ts
```diff
@@ -9,3 +10,27 @@ export const checkIsManualOptions = (
...
+  openForms: {code: string; service: string};
+} => {
+  return component.openForms.dataSrc === 'referentielijsten';
```
sergei-maertens, line 23: I think you can re-use the `ComponentWithReferentielijsten` type again here:

```ts
()component: RadioComponentSchema): component is RadioComponentSchema & ComponentWithReferentielijsten => {...}
```

src/registry/radio/helpers.ts
```diff
@@ -9,3 +10,27 @@ export const checkIsManualOptions = (
...
+};
+
+// A type guard is needed because TS cannot figure out it's a discriminated union
+// when the discriminator is nested.
+// See https://github.com/microsoft/TypeScript/issues/18758
+export const checkIsVariableOptions = (
+  component: RadioComponentSchema
+): component is RadioComponentSchema & {
+  data: {values: Option[] | undefined};
```
sergei-maertens, line 32:
```suggestion
```
same as above - the `values` will not be set since those are sorted out by the backend dynamically (which effectively transforms the component into dataSrc "manual")

src/registry/radio/radio-referentielijsten.stories.ts
```diff
@@ -0,0 +1,190 @@
...
+      type: 'radio',
+      key: 'radio',
+      label: 'A radio field',
+      dataSrc: 'values',
+      dataType: 'string',
+      openForms: {
+        dataSrc: 'manual',
+        translations: {},
+      },
+      data: {values: [{value: '', label: ''}]},
```
sergei-maertens, line 28:
```suggestion
```
this doesn't make sense for the radio component, does it?

src/registry/radio/radio-referentielijsten.stories.ts
```diff
@@ -0,0 +1,190 @@
...
+      const serviceInput = canvas.getByLabelText('Referentielijsten service');
+      await rsSelect(canvas, serviceInput, 'Referentielijsten');
+
+      const codeInput = canvas.getByLabelText('Referentielijsten table code');
+      await rsSelect(canvas, codeInput, 'Tabel 2 (niet meer geldig)');
+
+      await userEvent.click(canvas.getByRole('button', {name: 'Save'}));
+
+      expect(args.onSubmit).toHaveBeenCalledWith(
+        expect.objectContaining({
```
sergei-maertens, line 65: Today I learned!

src/registry/radio/radio-referentielijsten.stories.ts
```diff
@@ -0,0 +1,190 @@
...
+} as Meta<typeof ComponentEditForm>;
+
+type Story = StoryObj<typeof ComponentEditForm>;
+
+export const StoreValuesInComponent: Story = {
+  name: 'On save: store proper values in the component',
+  play: async ({canvasElement, step, args}) => {
+    const canvas = within(canvasElement);
+
+    await step('Fill in options', async () => {
```
sergei-maertens, line 52: can you drop the `step`s if there's only one step in the test?

src/registry/radio/radio-referentielijsten.stories.ts
```diff
@@ -0,0 +1,190 @@
...
+      const dataSourceInput = canvas.getByLabelText('Data source');
+      await rsSelect(canvas, dataSourceInput, 'Referentielijsten API');
+
+      const serviceInput = canvas.getByLabelText('Referentielijsten service');
+
+      await waitFor(async () => {
```
sergei-maertens, line 185:
```suggestion
      await waitFor(() => {
```
there's only sync code inside :)

src/registry/select/edit-validation.ts
```diff
@@ -18,7 +18,11 @@ const buildValuesSchema = (intl: IntlShape) =>
       values: optionSchema(intl).array().min(1).optional(),
     }),
     openForms: z.object({
-      dataSrc: z.union([z.literal('manual'), z.literal('variable')]),
+      dataSrc: z.union([
+        z.literal('manual'),
+        z.literal('variable'),
+        z.literal('referentielijsten'),
+      ]),
```
sergei-maertens, line 25: same remark about missing `code` and `service` (similar to the `itemsExpression`)

sergei-maertens (changes requested):
the comments on stories/implementation for radio obviously also apply to select and selectboxes :)

src/utils/storybookTestHelpers.d.ts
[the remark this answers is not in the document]
stevenbal: hmm I'm not sure anymore, I think storybook complained at some point, but it seems to work fine if I remove it

src/registry/radio/helpers.ts
stevenbal: should I then still leave the `return component.openForms.dataSrc === 'referentielijsten';` part in there? It feels like both lines are stating the same thing then :thinking:

src/components/formio/select.tsx
stevenbal: I tried this initially with valueProperty="slug" which I thought was causing issues, but tried it again just now and it works, so I've removed the transformation. I do need this or some other type for the `select/preview.tsx` options though

src/components/builder/values/values-config.stories.tsx
stevenbal: Adding `decorators` to the default story seems to cause errors: https://github.com/open-formulieren/formio-builder/actions/runs/13134517347/job/36646878514?pr=204#step:7:2152

src/registry/radio/helpers.ts
sergei-maertens: yeah, that's how type guards work :) if the outcome of the runtime check is `true`, then the `component is FOO` thing applies (which narrows the type down to this particular discriminated union member). If it's `false`, all you know is that it's a `RadioComponentSchema` which can still be any of the other union members.

tsconfig.json
```diff
@@ -16,6 +16,7 @@
     "noImplicitAny": true,
     "strictBindCallApply": true,
     "strictNullChecks": true,
+    "skipLibCheck": true,
```
sergei-maertens, line 19: Can you create a follow up issue to enable this again and assign it to me? I don't want to forget about it.

sergei-maertens (approved):

tsconfig.json
stevenbal: https://github.com/open-formulieren/formio-builder/issues/211

src/registry/radio/helpers.ts
stevenbal: ah like that, I was reading it the other way around, didn't know that the part after `:` was dependent on the return value, thanks for the explanation :+1:

PR #206, opened 2025-01-16
[OF#5006] Update addressNL component with `autoPopulated` field

vaszig:
Closes open-formulieren/open-forms#5006 partly

Requires: 

- [x] open-formulieren/types#65

### DIFF
--- a/src/components/ComponentConfiguration.stories.tsx
+++ b/src/components/ComponentConfiguration.stories.tsx
@@ -2360,6 +2360,9 @@ export const AddressNL: Story = {
         required: false,
       },
       deriveAddress: false,
+      defaultValue: {
+        autoPopulated: false,
+      },
       layout: 'singleColumn',
       openForms: {
         components: {
--- a/src/components/ComponentPreview.stories.tsx
+++ b/src/components/ComponentPreview.stories.tsx
@@ -1055,6 +1055,9 @@ export const AddressNL: Story = {
       },
       deriveAddress: false,
       layout: 'singleColumn',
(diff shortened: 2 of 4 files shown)


src/components/ComponentConfiguration.stories.tsx
```diff
@@ -2360,6 +2360,7 @@ export const AddressNL: Story = {
         required: false,
       },
       deriveAddress: false,
+      autoPopulated: false,
```
robinmolen, line 2363: This should be a part of the `defaultValue`, right?
```suggestion
    defaultValue: {
        autoPopulated: false,
    },

File: src/components/ComponentPreview.stories.tsx [hunk 2]
```diff
@@ -1055,6 +1055,7 @@ export const AddressNL: Story = {
       },
       deriveAddress: false,
       layout: 'singleColumn',
+      autoPopulated: false,
```
robinmolen, line 1058:
```suggestion
    defaultValue: {
        autoPopulated: false,
    },

robinmolen (approved):
It looks good to me 👍

sergei-maertens (commented):
Some thoughts here as in https://github.com/open-formulieren/types/pull/65#pullrequestreview-2572161224 and I don't see which changes this makes in the form builder - there don't appear to be any story snapshot changes in Chromatic?

edit: https://github.com/open-formulieren/open-forms/pull/5022/files#r1928413221 clarifies some things, dismissing this

PR #213, opened 2025-03-07
[OF#5107] Added referentielijsten options to the component's preview

vaszig:
Closes open-formulieren/open-forms#5107 partly


src/components/ComponentConfiguration.stories.tsx
```diff
@@ -1610,7 +1616,7 @@ export const Select: Story = {
       previewSearchInput.focus();
       await userEvent.keyboard('[ArrowDown]');
       await preview.findByRole('listbox');
-      await expect(await preview.findByRole('option', {name: 'Second option'})).toBeVisible();
+      await expect(preview.findByRole('option', {name: 'Second option'})).resolves.toBeVisible();
```
sergei-maertens, line 1619: that doesn't seem right, but I suppose you're having flakiness here? Can't we instead set a longer timeout for the `findByRole` or wrap it in an `await waitFor(...)`?

src/components/ComponentConfiguration.stories.tsx
```diff
@@ -7,7 +7,10 @@ import {Meta, StoryFn, StoryObj} from '@storybook/react';
 import {expect, fireEvent, fn, userEvent, waitFor, within} from '@storybook/test';
 import React from 'react';
 
-import {ReferentielijstenServiceOption} from '@/components/builder/values/referentielijsten/service';
+import {
+  ReferentielijstenServiceOption,
+  ReferentielijstenTabelItem,
+} from '@/components/builder/values/referentielijsten/types';
```
sergei-maertens, line 13:
```suggestion
import {
  ReferentielijstenServiceOption,
  ReferentielijstenTabelItem,
} from './builder/values/referentielijsten/types';
```
I generally prefer the explicit `@/` imports, but this file already has a bunch of imports so let's keep a consistent style within this file

src/components/builder/values/referentielijsten/code.tsx
```diff
@@ -3,45 +3,18 @@ import {useContext} from 'react';
 import {FormattedMessage, useIntl} from 'react-intl';
 import useAsync from 'react-use/esm/useAsync';
 
+import {transformItems} from '@/components/builder/values/referentielijsten/utils';
```
sergei-maertens, line 6:
```suggestion
import {transformItems} from './utils';
```

src/components/builder/values/referentielijsten/utils.ts
```diff
@@ -0,0 +1,25 @@
...
+  return items.map(item => {
+    const {code, naam, isGeldig} = item;
+    return {
+      value: code,
+      label: !isGeldig
+        ? `${naam} ${intl.formatMessage({
+            description: 'Message to indicate that Referentielijsten tabel is expired',
+            defaultMessage: '(niet meer geldig)',
+          })}`
+        : naam,
```
sergei-maertens, line 22: The correct way would be to incorporate the label directly in the message so that order of sentences can be adapted if needed:

```suggestion
      label: intl.formatMessage({
        description: 'Label for referentielijsten table or item',
        defaultMessage: '{naam}{isValid, select, true {} other { (no longer valid)}}',
      }, {name: naam, isValid: isGeldig})
```
Reference for message syntax: https://formatjs.github.io/docs/core-concepts/icu-syntax/

src/components/formio/radio.tsx
```diff
@@ -65,44 +67,55 @@ export const Radio: React.FC<RadioProps> = ({
...
-              <RadioInput
-                name={name}
-                value={value}
-                label={label || <EmptyLabel />}
-                description={description}
-              />
-            </label>
+    <>
+      {isLoading ? (
+        intl.formatMessage({
```
sergei-maertens, line 80: please use `<FormattedMessage ... />` if you can render normal React components. `intl.formatMessage` is a fallback for when you can't return a react component.

sergei-maertens, line 80: instead of doing the ternary operator, can't you simply put a guard clause before the existing return?

```jsx
if (isLoading) {
    return <FormattedMessage ... />;
}
return (<Component>...</Component>);
```

src/components/formio/selectboxes.tsx
```diff
@@ -24,25 +27,42 @@ export const SelectBoxes: React.FC<SelectBoxesProps> = ({
...
-                label={label}
-                optionDescription={description}
-              />
-            </label>
+    <>
+      {isLoading ? (
+        intl.formatMessage({
+          description: 'Text for loading values',
+          defaultMessage: 'Loading values...',
+        })
```
sergei-maertens, line 39: Same remarks as for radio component

src/registry/radio/preview.tsx
```diff
@@ -22,36 +26,47 @@ const Preview: React.FC<ComponentPreviewProps<RadioComponentSchema>> = ({compone
...
+              expression: JSON.stringify(component.openForms.itemsExpression),
+              code: chunks => <code>{chunks}</code>,
+            }
+          ),
+        },
+      ];
+    }
+
+    if (checkIsReferentielijstenOptions(component)) {
+      const response: ReferentielijstenTabelItem[] = await getReferentielijstenTabelItems(
```
sergei-maertens, line 58:
```suggestion
      const items: ReferentielijstenTabelItem[] = await getReferentielijstenTabelItems(
```
`response` implies it's an HTTP/API response, but the return type does not agree with that.

src/registry/select/preview.tsx
```diff
@@ -22,36 +26,47 @@ const Preview: React.FC<ComponentPreviewProps<SelectComponentSchema>> = ({compon
...
+              expression: JSON.stringify(component.openForms.itemsExpression),
+              code: chunks => <code>{chunks}</code>,
+            }
+          ),
+        },
+      ];
+    }
+
+    if (checkIsReferentielijstenOptions(component)) {
+      const response: ReferentielijstenTabelItem[] = await getReferentielijstenTabelItems(
```
sergei-maertens, line 58: see earlier remark

src/registry/selectboxes/preview.tsx
```diff
@@ -22,36 +26,45 @@ const Preview: React.FC<ComponentPreviewProps<SelectboxesComponentSchema>> = ({c
...
+            {
+              expression: JSON.stringify(component.openForms.itemsExpression),
+              code: chunks => <code>{chunks}</code>,
+            }
+          ),
+        },
+      ];
+    }
+    if (checkIsReferentielijstenOptions(component)) {
+      const response: ReferentielijstenTabelItem[] = await getReferentielijstenTabelItems(
```
sergei-maertens, line 57: see earlier remark

src/tests/sharedUtils.tsx
```diff
@@ -65,6 +69,21 @@ export const DEFAULT_REFERENTIELIJSTEN_TABELLEN: ReferentielijstenTabelOption[]
...
+      code: 'option1',
+      naam: 'Option 1',
+      isGeldig: true,
+    },
+    {
+      code: 'option2',
+      naam: 'Option 2',
+      isGeldig: true,
+    },
+  ],
```
sergei-maertens, line 84: Let's use some real data :)

```suggestion
  countries: [
    {
      code: 'nl',
      naam: 'The Netherlands',
      isGeldig: true,
    },
    {
      code: 'gr',
      naam: 'Greece',
      isGeldig: true,
    },
  ],
```

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #214, opened 2025-03-11
Removing zIndex from required field asterisks

robinmolen:
Closes open-formulieren/open-forms#5105

~The asterisks on required fields have a z-index of 10. This is part of base formio styling, so no idea why this is. Instead of lowering this, I've opted to raise the select component menu.~

Edit: So I've changed this to indeed remove the z-index from the required field asterisks. Raising the select menu is also an option, but that scares me a bit more

### DIFF
--- a/src/components/formio/component-label.scss
+++ b/src/components/formio/component-label.scss
@@ -1,0 +1,5 @@
+.field-required {
+  &:after {
+    z-index: unset;
+  }
+}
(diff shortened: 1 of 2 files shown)


robinmolen:
Okay, took a second look and found the PR that added the `z-index: 10px;` to the formio styling https://github.com/formio/formio.js/pull/3692. Maybe we can safely remove this, as (at least, i think so) all our fields have their labels visible.

sergei-maertens (approved):
Seems to do the job indeed - however when we integrate it in the open forms backend we'll have to be careful with the order of stylesheets imported - if we import the builder CSS after the formio CSS we should be good, if the order is the other way around, our rule will be overridden again by Formio.

PR #218, opened 2025-03-24
[OF#5178] Do not show expired tabels and table items in the options (referentielijsten)

### LINKED ISSUE #5178: Expired items should not show up in the list (populated from reference lists) and the admin should be notified if all items expired.
### Thema / Theme

Frontend

### Omschrijving / Description

...or if the table expired.

* [ ] Expired items should not show up in the list (public)
* [ ] If all items expired or the table expired, the admin should get a message in digest email

### Added value / Toegevoegde waarde

_No response_

### Aanvullende opmerkingen / Additional context

_No response_

vaszig:
Closes open-formulieren/open-forms#5178 partly

### DIFF
--- a/src/components/ComponentConfiguration.stories.tsx
+++ b/src/components/ComponentConfiguration.stories.tsx
@@ -1134,7 +1134,32 @@ export const SelectBoxes: Story = {
       values: [],
       defaultValue: {},
     },
-
+    referenceListsTables: [
+      {
+        code: 'countries',
+        naam: 'Countries',
+        isGeldig: true,
+      },
+      {
+        code: 'invalid-table',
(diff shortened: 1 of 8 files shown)


sergei-maertens (commented):
Err, I'd expect that we *do* show the expired items in the form builder, but not in the SDK/backend so that it's at least clear that the items have expired, no?

vaszig:
> Err, I'd expect that we _do_ show the expired items in the form builder, but not in the SDK/backend so that it's at least clear that the items have expired, no?

Ok, I will close this PR then. I understood that we do not want to show the expired generally and  the admin users will be informed by the digest email.

PR #223, opened 2025-04-10
Catching component preview errors

robinmolen:
Partly closes open-formulieren/open-forms#5191

Adding a ErrorBoundary component, with generic error handling, to make sure that errors thrown in the component preview don't crash the component modal.

### DIFF
--- a/src/components/ComponentPreview.tsx
+++ b/src/components/ComponentPreview.tsx
@@ -4,10 +4,11 @@ import {Formik} from 'formik';
 import React, {useContext, useState} from 'react';
 import {FormattedMessage} from 'react-intl';
 
-import PreviewModeToggle, {PreviewState} from '@/components/PreviewModeToggle';
+import PreviewModeToggle, {type PreviewState} from '@/components/PreviewModeToggle';
+import ErrorBoundary from '@/components/error/ErrorBoundary';
 import {BuilderContext} from '@/context';
 import {getRegistryEntry} from '@/registry';
-import {AnyComponentSchema, hasOwnProperty} from '@/types';
+import {type AnyComponentSchema, hasOwnProperty} from '@/types';
 
 /*
   Generic preview (preview + wrapper with view mode)
@@ -60,7 +61,7 @@ const ComponentPreviewWrapper: React.FC<ComponentPreviewWrapperProps> = ({
               className={clsx('component-preview', `component-preview--${component.type}`)}
               data-testid="componentPreview"
             >
-              {children}
+              <ErrorBoundary>{children}</ErrorBoundary>
             </div>
           </Formik>
         </div>
--- a/src/components/error/ErrorBoundary.stories.tsx
+++ b/src/components/error/ErrorBoundary.stories.tsx
@@ -1,26 +1,19 @@
-import {Meta, StoryObj} from '@storybook/react';
+import type {Meta, StoryObj} from '@storybook/react';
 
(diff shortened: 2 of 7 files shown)


src/components/ComponentPreview.tsx
```diff
@@ -32,40 +33,42 @@ const ComponentPreviewWrapper: React.FC<ComponentPreviewWrapperProps> = ({
...
-            </div>
-          </Formik>
-        </div>
-      )}
-    </div>
+              <div
+                className={clsx('component-preview', `component-preview--${component.type}`)}
+                data-testid="componentPreview"
+              >
+                {children}
```
sergei-maertens, line 65: I'd restrict the error boundary to this node:

* bugs in our own JSONEditor or Formik setup should not be suppressed, we need to be aware of them
* crashes in the preview will still allow you to fix the JSON if needed, that's currently not possible because the preview mode toggle is in the error boundary
* only the `{children}` our outside of our own control (they come from the registry)

src/components/ErrorBoundary.stories.tsx
sergei-maertens, line 1: The SDK TS improvements for this have just been merged today, let's copy that version instead. https://github.com/open-formulieren/open-forms-sdk/blob/main/src/components/Errors/ErrorBoundary.stories.tsx

sergei-maertens (changes requested):
One possible additional improvement is to update the map preview component and run the Zod schema validation on the `component` that's passed in for preview. If it doesn't validate, we can display an informative message saying that the configuration is invalid, rather than crashing. (but keep the error boundary too, other stuff can break of course)

robinmolen:
I've updated the `ErrorBoundary` component and moved the separate components to separate files (making it easier to add new/other error types).

I'll take a look at using Zod to validate the map data

sergei-maertens:
> I've updated the `ErrorBoundary` component and moved the separate components to separate files (making it easier to add new/other error types).
> 
> I'll take a look at using Zod to validate the map data

@robinmolen zod is already set up, see https://github.com/open-formulieren/formio-builder/blob/main/src/registry/map/edit-validation.ts

robinmolen:
I have a setup for the Zod validation of the map preview, but will put that in a separate PR. This way we can merge this PR "fast", while fine tuning the Zod validation part

src/components/error/ErrorBoundary.stories.tsx
sergei-maertens, line 1: in the SDK we have this:

```tsx
// in JS, you actually *can* throw anything
const Throw: React.FC<{error: AnyError}> = ({error}) => {
  throw error;
};

export default {
  title: 'Private API / ErrorBoundary',
  component: ErrorBoundary,
  args: {
    children: <Throw error={new Error('nope')} />,
    useCard: true,
  },
  argTypes: {
    children: {table: {disable: true}},
  },
} satisfies Meta<typeof ErrorBoundary>;
```

which is a bit nicer since you don't need the custom `render` function

src/components/error/ErrorBoundary.stories.tsx
```diff
@@ -0,0 +1,29 @@
+import {Meta, StoryObj} from '@storybook/react';
```
sergei-maertens, line 1:
```suggestion
import type {Meta, StoryObj} from '@storybook/react';
```

sergei-maertens (changes requested):

PR #228, opened 2025-05-21
[#225] Add partners custom component

### LINKED ISSUE #225: Add custom 'partners' component type
This requires adding the new type definitions to https://github.com/open-formulieren/types and adding the component to the builder. Finally, a custom preview needs to be rendered in the backend too.

Example component definition:

```json
{
  "type": "partners",
  "key": "partners",
  "label": "Partners",
  "populateFrom": "partnersPrefill"
}
```

<details>
<summary>Wireframe</summary>

```
+-----------------------------+--------------------------------+
| Partners component          | Preview                      X |
|                             |                                |
|                             |                                |
| Label                       | BSN                            |
| ________________            | _________                      |
|                             |                                |
| Key                         | Initials                       |
| ________________            | ---------                      |
|                             |                                |
| Prefilled variable          | Affixes                        |
| +--------------------+      | ---------                      |
| | partnersPrefill  v |      |                                |
| +--------------------+      | Last name                      |
|                             | _________                      |
|                             |                                |
|                             | Date of birth                  |
|                             | _________                      |
|                             |                                |
|                             |                                |
|                             | +--------+ +--------+ +------+ |
|                             | | Cancel | | Remove | | Save | |
|                             | +--------+ +--------+ +------+ |
+--------------------------------------------------------------+
```

</details>

**Acceptance criteria**

- [x] A new component type "partners" is available in the formio-builder storybook
- [ ] In the backend, I can add a partners component to the form. The preview panel gives an impression of the layout/structure in the SDK and the full-form preview does so too.

**Tasks**

- [x] Define TS type definition in https://github.com/open-formulieren/types
- [x] Implement builder configuration form in formio-builder & add stories
- [x] Add dropdown for available form variables to select (for `populateFrom`)
- [ ] Bonus: check if we can filter the available variables above to the array variables configured with appropriate prefill plugin. This filtering should ideally be performed in the backend, since the formio-builder doesn't know the "meaning" of prefill plugins (and it shouldn't).
- [x] Enable the custom component type in the backend form builder
    - Implement a minimal custom component type and render preview. You can look at addressNL for inspiration.

vaszig:
Closes #225 partly

- [x] Requires open-formulieren/types#70

### DIFF
--- a/src/components/ComponentConfiguration.stories.tsx
+++ b/src/components/ComponentConfiguration.stories.tsx
@@ -3059,3 +3059,47 @@ export const SoftRequiredErrors: Story = {
     });
   },
 };
+
+export const Partners: Story = {
+  render: Template,
+  name: 'type: partners',
+
+  args: {
+    component: {
+      id: 'wekruya',
+      type: 'partners',
+      key: 'partners',
+      label: 'Partners',
+      tooltip: 'An example for the tooltip',
+      description: 'A description for the Partners component',
+      defaultValue: [],
+    },
+  },
+
+  play: async ({canvasElement, step, args}) => {
+    const canvas = within(canvasElement);
+
+    expect(canvas.getByLabelText('Label')).toHaveValue('Partners');
+    await waitFor(() => {
+      expect(canvas.getByLabelText('Property Name')).toHaveValue('partners');
+    });
+    expect(canvas.getByLabelText('Is sensitive data')).toBeChecked();
+
+    await step('Change key', async () => {
+      // Ensure that the manually entered key is kept instead of derived from the label,
+      // even when key/label components are not mounted.
+      const keyInput = canvas.getByLabelText('Property Name');
+      fireEvent.change(keyInput, {target: {value: 'customKey'}});
+      await userEvent.click(canvas.getByRole('tab', {name: 'Translations'}));
+      await userEvent.click(canvas.getByRole('tab', {name: 'Basic'}));
+      await userEvent.clear(canvas.getByLabelText('Label'));
+      await userEvent.type(canvas.getByLabelText('Label'), 'Other label', {delay: 50});
+      expect(canvas.getByLabelText('Property Name')).toHaveDisplayValue('customKey');
+    });
+
+    await step('Submit form', async () => {
+      await userEvent.click(canvas.getByRole('button', {name: 'Save'}));
+      expect(args.onSubmit).toHaveBeenCalled();
+    });
+  },
+};
--- a/src/components/ComponentPreview.stories.tsx
+++ b/src/components/ComponentPreview.stories.tsx
@@ -5,6 +5,7 @@ import {
   EditGridComponentSchema,
   FieldsetComponentSchema,
   FileComponentSchema,
+  PartnersComponentSchema,
   PhoneNumberComponentSchema,
   PostcodeComponentSchema,
   RadioComponentSchema,
@@ -1328,5 +1329,20 @@ export const LeafletMap: Story = {
     // the map should be rendered, with zoom controls
     await expect(await canvas.findByRole('button', {name: 'Zoom in'})).toBeVisible();
     await expect(await canvas.findByRole('button', {name: 'Zoom out'})).toBeVisible();
+  },
+};
+
+export const Partners: Story = {
+  name: 'Partners',
+  args: {
+    component: {
+      id: 'wekruya',
+      type: 'partners',
+      key: 'partners',
+      label: 'A partners preview',
+      tooltip: 'An example for the tooltip',
+      description: 'A description for the Partners component',
+      defaultValue: [],
+    } satisfies PartnersComponentSchema,
   },
 };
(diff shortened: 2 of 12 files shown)


src/components/builder/prefill/family-members/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+  const tooltip = intl.formatMessage({
+    description: "Tooltip for 'populateFrom' builder field",
+    defaultMessage:
+      'Select the user defined form variable from which this component will be prefilled.',
+  });
+
+  const availableFormVariables: JSONObject[] = getAvailableFormVariables();
+
+  // We only need to show the relevant variables that can be used for the Partners component
+  const _options: JSONObject[] = availableFormVariables
```
vaszig, line 27: In the task there is a bonus subtask which would filter the variables in the backend. I chose to do it here instead because I thought that the callback should be generic (just retrieve all the available variables) and do the special handling in the specific case where it's used. I would like your opinion.

src/components/ComponentConfiguration.stories.tsx
```diff
@@ -3059,3 +3059,55 @@ export const SoftRequiredErrors: Story = {
...
+
+  args: {
+    component: {
+      id: 'wekruya',
+      type: 'partners',
+      key: 'partners',
+      label: 'Partners',
+      tooltip: 'An example for the tooltip',
+      description: 'A description for the Partners component',
+      defaultValue: [
```
sergei-maertens, line 3075: does it make sense to write stories with `defaultValue` when we don't allow this to be set?

src/registry/partners/edit.tsx
```diff
@@ -0,0 +1,104 @@
...
+        <BuilderTabs.Basic
+          hasErrors={hasAnyError(
+            'label',
+            'key',
+            'description',
+            'tooltip',
+            'hidden',
+            'clearOnHide',
+            'isSensitiveData',
+            'defaultValue'
```
sergei-maertens, line 44: this is not a field that can be edited, so tracking validation errors for it makes little sense

src/registry/partners/edit.tsx
```diff
@@ -0,0 +1,104 @@
...
+  isSensitiveData: true,
+  defaultValue: [
+    {
+      bsn: '',
+      initials: '',
+      affixes: '',
+      lastName: '',
+      dateOfBirth: '',
+    },
+  ],
```
sergei-maertens, line 101:
```suggestion
  defaultValue: [],
```
even though it will be overwritten by the backend, I'd go for an empty list here and set up some dummy values in the preview component instead

src/components/builder/is-sensitive-data.tsx
```diff
@@ -19,6 +19,7 @@ const IsSensitiveData = () => {
         />
       }
       tooltip={tooltip}
+      disabled={disabled}
```
sergei-maertens, line 22: I suspect you mean to use `readonly` here rather than `disabled`. With the Formik state, there's no difference, but in pure HTML semantics a `disabled` form input is **NOT** submitted to the server, while a readonly input is.

The value of the property/field is important here for us, so I'd rather go for readonly than disabled.

Formio itself also famously gets this wrong where they use the `disabled` property when they actually mean `readonly`.

src/components/ComponentPreview.stories.tsx
```diff
@@ -1330,3 +1331,26 @@ export const LeafletMap: Story = {
...
+      description: 'A description for the Partners component',
+      defaultValue: [
+        {
+          bsn: '123456788',
+          initials: 'J',
+          affixes: 'Van',
+          lastName: 'Deniston',
+          dateOfBirth: '2000-09-08',
+        },
+      ],
```
sergei-maertens, line 1353:
```suggestion
      defaultValue: [],
```
I think the preview should make sense even when "nothing" is configured on the component.

We make similar pragmatic choices like that in the columns and edit grid component where we just render some placeholder data instead of "real" children/default values.

src/registry/partners/index.ts
```diff
@@ -0,0 +1,23 @@
...
+  },
+  defaultValue: [
+    {
+      bsn: '',
+      initials: '',
+      affixes: '',
+      lastName: '',
+      dateOfBirth: '',
+    },
+  ],
```
sergei-maertens, line 22:
```suggestion
  defaultValue: [],
```

src/registry/partners/preview.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+const Preview: React.FC<ComponentPreviewProps<PartnersComponentSchema>> = ({component}) => {
+  const {key, label, description, tooltip, defaultValue = []} = component;
+
+  return (
+    <FieldSet field={key} label={label} tooltip={tooltip}>
+      <TextField
+        name={`${key}.bsn`}
+        label={<FormattedMessage description="Label for partner's BSN" defaultMessage="BSN" />}
+        value={defaultValue[0].bsn}
```
sergei-maertens, line 17: you may use fictive hardcoded preview data here instead of looking into `defaultValue`

sergei-maertens (changes requested):
Keep in mind that when you specify a `defaultValue`, it actually gets saved in the JSON of the component definition and thus in the database eventually. To avoid confusion, I think it's best to leave `defaultValue` empty.

src/components/builder/is-sensitive-data.tsx
vaszig: Nice, I didn't know about this important difference.

PR #239, opened 2025-09-25
Catch validation errors in map component preview

robinmolen:
Partly closes https://github.com/open-formulieren/open-forms#5191

This PR is an extension on top of https://github.com/open-formulieren/formio-builder/pull/223. By validation the map component in the map preview, we can highlight any configuration issues gracefully instead of risking hard crashes.

The ErrorBoundary now supports Zod errors. The Zod errors use a different "error message" template, highlighting the invalid configuration fields and their issues. Currently this shows the internal field name (like "lat" for "latitude").

### DIFF
--- a/i18n/messages/en.json
+++ b/i18n/messages/en.json
@@ -124,6 +124,11 @@
     "description": "Description for 'fileMaxSize' builder field",
     "originalDefault": "Note that the server upload limit is {serverUploadLimit}."
   },
+  "2gs8pJ": {
+    "defaultMessage": "The map configuration is not valid, so we can't show a preview. Fix the validation errors in the component configuration.",
+    "description": "Map preview: invalid config message",
+    "originalDefault": "The map configuration is not valid, so we can't show a preview. Fix the validation errors in the component configuration."
+  },
   "2k+Mca": {
     "defaultMessage": "Validation",
     "description": "Component edit form tab title for 'Validation' tab",
--- a/i18n/messages/nl.json
+++ b/i18n/messages/nl.json
@@ -125,6 +125,11 @@
     "description": "Description for 'fileMaxSize' builder field",
     "originalDefault": "Note that the server upload limit is {serverUploadLimit}."
   },
+  "2gs8pJ": {
+    "defaultMessage": "The map configuration is not valid, so we can't show a preview. Fix the validation errors in the component configuration.",
+    "description": "Map preview: invalid config message",
+    "originalDefault": "The map configuration is not valid, so we can't show a preview. Fix the validation errors in the component configuration."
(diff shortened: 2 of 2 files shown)


robinmolen:
@sergei-maertens I'm not sure about showing the Zod details (or at least, how i've implemented at this moment).

At the one hand, I think its best to show some details (so we don't get situations of "the form says there is an error, but i don't see it"). On the other hand, I don't know how we can nicely relate the internal field name to the field label, and the internal names might just make things more complicated/confusing.

src/registry/map/preview.tsx
```diff
@@ -94,6 +99,14 @@ const Preview: React.FC<ComponentPreviewProps<MapComponentSchema>> = ({component
     return <Loader />;
   }
 
+  try {
+    zodSchema({intl, builderContext}).parse(component);
+  } catch (error) {
+    if (error instanceof z.ZodError) {
+      throw error;
+    }
+  }
```
sergei-maertens, line 108:
```suggestion
  const {success: isValidConfiguration} = zodSchema({intl, builderContext}).safeParse(component)
  
  if (!isValidConfiguration) {
    return <FormattedMessage description="Map preview: invalid config message" defaultMessage="The map configuration is not valid, so we can't show a preview. Fix the validation errors in the component configuration." />;
  };
```
Something like that :)

sergei-maertens (changes requested):
I had a much simpler approach in mind tbh 😬 

I think all the changes to the error boundary itself can be reverted:

* the property vs. label confusion you raise yourself is not ideal
* I don't like the complexity this brings

Instead, I would just do the following:

* in the map preview component, call the schema validation, see the code suggestion
* don't throw, but just display a generic error message saying things are invalid and to look at  the configuration field validation errors

src/registry/map/map-configuration.stories.ts
```diff
@@ -18,6 +18,9 @@ export default {
   decorators: [BuilderContextDecorator],
   parameters: {
     builder: {enableContext: true},
+    test: {
+      dangerouslyIgnoreUnhandledErrors: true,
+    },
```
sergei-maertens, line 23: why do we need this?

robinmolen: Ah snap, yeah that isn't needed anymore...

The errorboundary logs all errors (so we show the generic error message, and provide additional context in the browser log). With the previous setup, of Zod also being caught by the error boundary, there would be expected error logs of the Zod validation errors. These error logs would cause the storybook tests to fail, as they were seen as unhandled errors.

sergei-maertens (approved):

PR #252, opened 2025-12-15
Bug/5723 map overlay configuration automatic closing

robinmolen:
closes https://github.com/open-formulieren/open-forms/issues/5723

This solves the problem of map overlay configuration automatically closing when changing the overlay tile or title. The problem here was that the array index, overlay tile and overlay title were combined to create a unique component key. Which every value change, the key would update, triggering an re-render and closing the configuration menu.

The solution here is to add an internal identifier to the overlay data, which helps with keeping track of which overlay is which. This solution adds the internal identifier to the overlay data, which is later submitted to the backend.

I want to prevent sending this data to the backend, but im not sure how to best approach this. Should this data be removed before submission? Should we store this information somewhere else? Should the backend remove this information via the serializers?

### DIFF
--- a/src/registry/map/map-configuration.stories.ts
+++ b/src/registry/map/map-configuration.stories.ts
@@ -53,7 +53,7 @@ export const NotUsingGlobalConfig: Story = {
     });
 
     await step('Open configuration panel', async () => {
-      const panelTitle = await canvas.findByText('Initial focus');
+      const panelTitle = await canvas.findByRole('button', {name: 'Initial focus'});
       await waitFor(async () => {
         expect(panelTitle).toBeVisible();
       });
@@ -260,7 +260,9 @@ export const EditOverlay: Story = {
     const overlayConfigTrigger = await canvas.findByRole('button', {
       name: 'Overlay: BAG pand layer',
     });
-    expect(overlayConfigTrigger).toBeVisible();
+    await waitFor(() => {
+      expect(overlayConfigTrigger).toBeVisible();
+    });
     await userEvent.click(overlayConfigTrigger);
 
     await step('Validate initial values', async () => {


sergei-maertens:
> I want to prevent sending this data to the backend, but im not sure how to best approach this. Should this data be removed before submission? Should we store this information somewhere else? Should the backend remove this information via the serializers?

It'll just be ignored by the backend eventually, don't fuss about it. No need to put in extra effort.

While it currently may be stored in the JSONField, my experiment with msgspec looks promising and one of the things it does is ignore unspecified fields.

src/components/ComponentConfiguration.stories.tsx
```diff
@@ -2933,9 +2935,95 @@ export const LeafletMap: Story = {
       await expect(canvas.getByLabelText('Property Name')).toHaveDisplayValue('customKey');
     });
 
+    await step('Set map settings', async () => {
+      await userEvent.click(canvas.getByRole('tab', {name: 'Map settings'}));
+      await expect(canvas.getByLabelText('Polygon')).not.toBeChecked();
```
sergei-maertens: `await expect` does not make sense, `expect` is always sync.

src/registry/map/edit.tsx
```diff
@@ -52,6 +53,24 @@ const EditForm: EditFormDefinition<MapComponentSchema> = () => {
...
+    const needsUpdate = overlays.some(o => !o._OF_INTERNAL_id);
+    if (!needsUpdate) return;
+
+    setValues({
+      ...values,
+      overlays: overlays.map(o => ({
+        ...o,
+        _OF_INTERNAL_id: o._OF_INTERNAL_id ?? crypto.randomUUID(),
+      })),
+    });
```
sergei-maertens: can't you `setValue` targetting just the `overlays` field? the object spread of `...values` breaks all references and causes a re-render of *every* field now, which is quite inefficient.

I also think that calling this effect only once is flawed, if someone pastes an old configuration in the JSON panel, this effect won't fire again and it re-introduces the problem. Perhaps this should be moved down into the component that renders a single layer panel and make it self-correct based on whether a unique ID is present or not.

src/registry/types.ts
```diff
@@ -49,6 +49,11 @@ export interface Previews {
   panel: React.FC<ComponentPreviewProps> | null;
 }
 
+/**
+ * Callback to update the component data before submission to backend.
+ */
+export type BeforeSubmitCallback<S extends AnyComponentSchema> = (data: S) => S;
```
sergei-maertens: Please revert this, it's so much overkill for a non-issue.

sergei-maertens (changes requested):

src/components/ComponentConfiguration.stories.tsx
```diff
@@ -2933,9 +2935,95 @@ export const LeafletMap: Story = {
       await expect(canvas.getByLabelText('Property Name')).toHaveDisplayValue('customKey');
     });
 
+    await step('Set map settings', async () => {
```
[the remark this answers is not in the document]
robinmolen: Reverting these changes

PR #261, opened 2026-03-12
bump storybook to 10

annashamray:
fixes https://github.com/open-formulieren/open-forms/issues/5991

**Changes:**
* bump storybook to 10  
* replace storybook test-runner with vitest
* replace jest with vitest
* add vite as a package builder

**what is left** 
* refactor unit tests and replace `@testing-library` packages

### DIFF
--- a/.storybook/main.ts
+++ b/.storybook/main.ts
@@ -1,4 +1,3 @@
-// This file has been automatically migrated to valid ESM format by Storybook.
 import type {StorybookConfig} from '@storybook/react-vite';
 
 const config: StorybookConfig = {
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -3,7 +3,7 @@
     "baseUrl": "src",
     "outDir": "lib/esm",
     "module": "esnext",
-    "target": "es2017",
+    "target": "ES2021",
     "jsx": "react-jsx",
     "declaration": true,
     "moduleResolution": "bundler",


vite.config.ts
```diff
@@ -63,4 +67,38 @@ export default defineConfig(({mode}) => ({
...
+      reporter: ['text', 'cobertura', 'html'],
+    },
+    browser: {
+      enabled: true,
+      headless: true,
+      provider: playwright({}),
+      instances: [{browser: 'chromium'}],
+      screenshotFailures: false,
+    },
+    projects: [
```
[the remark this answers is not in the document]
annashamray: Yes, my plan was to follow your advice and to make iterations as small as possible

package.json
```diff
@@ -19,10 +19,10 @@
     "build:esm": "tspc -p tsconfig.prod.json && tsc-alias -p tsconfig.prod.json",
     "build:cjs": "tspc -p tsconfig.prod.json --module commonjs --outDir lib/cjs && tsc-alias -p tsconfig.prod.json --outDir lib/cjs",
     "build:scss": "node bundle-scss.mjs && rm lib/css/index.js lib/css/index.js.map",
-    "build:typecheck": "tsc --noEmit",
+    "build:typecheck": "tsc --noEmit -p tsconfig.prod.json",
```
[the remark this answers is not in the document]
annashamray: I removed this workaround switching to vite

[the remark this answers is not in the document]
sergei-maertens: Vite is love, Vite is life

.github/workflows/ci.yml
```diff
@@ -86,38 +91,38 @@ jobs:
...
-  tests-jest:
-    name: Run Jest tests
+  unit-tests:
+    name: Run unit tests
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-node@v4
+      - uses: actions/checkout@v6
```
sergei-maertens: Can you pin these to commit hashes and add a comment to which tag they point? We're hardening our supply chain, see https://github.com/open-formulieren/open-forms/issues/6126

.storybook/main.ts
```diff
@@ -1,3 +1,4 @@
+// This file has been automatically migrated to valid ESM format by Storybook.
```
sergei-maertens:
```suggestion
```

src/components/formio/affix.tsx
```diff
@@ -1,17 +1,17 @@
...
+  ALLOWED_ATTR: [],
 };
 
 export interface AffixProps {
   children: string;
 }
 
 const Affix: React.FC<JSX.IntrinsicElements['span'] & AffixProps> = ({children, ...props}) => {
-  const sanitizedContent = sanitizeHtml(children, OPTIONS);
+  const sanitizedContent = DOMPurify.sanitize(children, OPTIONS);
```
sergei-maertens, line 13: it looks like (see Chromatic comment) the `strong` tags are still slipping through 🤔 

The documentation mentions:

```
// allow all safe HTML elements but neither SVG nor MathML
// note that the USE_PROFILES setting will override the ALLOWED_TAGS setting
// so don't use them together
```

package.json
```diff
@@ -85,46 +73,50 @@
...
     "vite-plugin-dts": "^4.5.4",
-    "vite-tsconfig-paths": "^5.1.4"
+    "vite-tsconfig-paths": "^5.1.4",
+    "vitest": "^4.0.18",
+    "yargs": "^18.0.0"
   },
   "dependencies": {
     "@ckeditor/ckeditor5-react": "^11.0.1",
     "@floating-ui/react": "^0.26.4",
+    "@formio/vanilla-text-mask": "^5.1.1",
```
sergei-maertens, line 108: Hrm, where is this coming from?

tsconfig.json
```diff
@@ -4,10 +4,9 @@
     "outDir": "lib/esm",
     "module": "esnext",
     "target": "es2017",
```
sergei-maertens: I think we can target somehing more recent, I believe we use 2021 in renderer/SDK but we can be a bit more aggressive even for the admin UI IMO

package.json
annashamray: it's `applyInputMask` which is used for textfield:
https://github.com/open-formulieren/formio-builder/blob/73bf6bb85e5ce4e61cf68fc9d1b0651a7eb1e258/src/utils/inputmask.ts#L16

sergei-maertens: right, in another PR/cleanup we'll be able to clean that up then since we don't support that `inputMask` feature in the new renderer anyway :D

.github/workflows/ci.yml
annashamray: pinned to commit hashes

.storybook/main.ts
annashamray: removed

src/components/formio/affix.tsx
annashamray: yeah it was probably not a good idea to use `USE_PROFILES`. Removed

tsconfig.json
annashamray: I'll go with the safe approach for now and switch to ES2021
Admin UI also use 2021 btw

PR #275, opened 2026-04-15
Implement components list

### LINKED ISSUE #266: Form-designer - implement components list
Implement the list of component elements available to the Form-designer. The components are grouped by 'common', 'special', 'advanced', ...
We use the existing functionality, plus some UI fixes (in the current Ui, filtering the list and then clearing the filter, disables the ability to open component group accordions)

This entails:
- [x] The displaying of component groups
- [x] Making the component groups collapsable/expandable
By clicking the group name you can open the group. This automatically closes the other group. At any time, only one group can be opend.
- [x] Showcasing the components within the groups (with icons and translatable labels)
- [x] Search input to filter the list of components

## Searching
When a search query is present, all groups are opened. Only the components that contain the search query in their name (technical name + translated names) are displayed. When a group has no matching components, that group gets hidden.
When the search query is cleared, all components and component groups are returned to the list.

robinmolen:
Closes #266

Implementing a ComponentsList component inside the formio-builder. This component closely resembles the current Formio implementation, and aims to replace that setup.

The ComponentsList component features collapsable component groups and search capabilities (also allowing searching on the translated labels, something the current implementation does not provide). The translations are done within the formio-builder project, to keep the control within the project.

The properties of the ComponentsList component might change in some way. Currently it takes the builder configuration and components structure as they would be available in the backend project. The actual implementation of this new form designer feature will probably have one entry point (an FormDesigner component that gets implemented in the backend project), which normalizes the data and passes that along to the ComponentsList. This final setup is still to be determined.


src/components/designer/ComponentsList.stories.tsx
```diff
@@ -0,0 +1,807 @@
...
+    docs: {
+      source: {
+        type: 'dynamic',
+        excludeDecorators: true,
+      },
+    },
+    modal: {noModal: true},
+  },
+  args: {
+    builder: {
```
sergei-maertens, line 20: I'm not at all married to this data structure, and since we control it ourselves, I think we can work with a simpler list of groups here perhaps?

src/components/designer/ComponentsList.stories.tsx
```diff
@@ -0,0 +1,807 @@
...
+          group: 'custom_special',
+          weight: 300,
+          schema: {},
+        },
+      },
+    },
+  },
+} as Meta<typeof ComponentsList>;
+
+export const Default: StoryObj<typeof ComponentsList> = {
```
sergei-maertens, line 569: You can break out the type:

```tsx
type Story = StoryObj<typeof ComponentsList>;
```
and re-use that simple type annotation - that matches the pattern we apply everywhere else :)

src/components/designer/ComponentsList.tsx
```diff
@@ -0,0 +1,142 @@
+import type {AnyComponentSchema} from '@open-formulieren/types';
+import clsx from 'clsx';
+import {useId, useMemo, useState} from 'react';
+import {useIntl} from 'react-intl';
+
+import './ComponentsList.scss';
+import {normalizeBuilderGroups} from './normalizeBuilderConfig';
+import type {BuilderGroup, ComponentDefinition, NormalizedBuilderGroup} from './types';
+
+interface ComponentsListProps {
```
sergei-maertens, line 10: I think you need to export this for auto-documentation features, and it seems like a good idea to have an `.mdx` file with documentation for this component?

src/components/designer/ComponentsList.tsx
```diff
@@ -0,0 +1,142 @@
...
+  const [searchQuery, setSearchQuery] = useState<string>('');
+  const intl = useIntl();
+  const defaultList = Object.entries(builder).find(
+    ([, config]) => config !== false && config.default
+  );
+  // Set the default list as initially expanded. If there is no default list, expand the
+  // first list.
+  const [expanded, setExpanded] = useState<string>(
+    defaultList ? defaultList[0] : builderListKeys[0]
+  );
```
sergei-maertens, line 36: can we use the native `details` elements for this possibly? The django admin fieldsets have also changed their divs + JS implementation to native HTML elements.

src/components/designer/ComponentsList.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+    // Search components by their key and title.
+    return normalizedGroups.map(group => ({
+      ...group,
+      components: group.components.filter(component => {
+        const searchTerms = [component.key.toLowerCase(), component.title.toLowerCase()];
+        return searchTerms.some(term => term.includes(searchQuery.toLowerCase()));
+      }),
+    }));
+  }, [normalizedGroups, searchQuery]);
```
sergei-maertens, line 58: I think this search behaviour is the prime example of the `useDeferredValue` hook in React!

The `useMemo` here seems like premature optimization? I'd expect the search query to be passed down as prop to the component that gets rendered for each group, and that determines whether the group itself should be expanded or not and filters the children of the group.

The react state changes then lead to prop changes and lets React re-render when necessary.

src/components/designer/normalizeBuilderConfig.ts
```diff
@@ -0,0 +1,77 @@
...
+
+import {COMPONENT_GROUP_LABELS, COMPONENT_TYPE_LABELS} from './messages';
+import type {
+  BuilderGroup,
+  ComponentBuilderInfo,
+  ComponentDefinition,
+  NormalizedBuilderGroup,
+} from './types';
+
+const getGroupLabel = (groupName: string, defaultTitle: string, intl: IntlShape) =>
```
sergei-maertens, line 12: why can't the `groupName` not be the `keyof COMPONENT_GROUP_LABELS` type? Why is a default title necessary?

sergei-maertens (changes requested):

src/components/designer/ComponentsList.stories.tsx
robinmolen: Mhh, i wanted to make it as easy as possible to implement in the backend, but fair point.

We can definitely change the dict to a list, and filter out the unused groups. Then there still is the matter of the components within the groups.

Currently the groups dict is normalized to a list containing the component details. Im not sure if this normalization should also be done in the backend. So perhaps just simplifying the groups structure to a list is enough

sergei-maertens: I think keeping the current structure is propagating technical debt rather than cleaning it up now that we have the chance :-)

src/components/designer/normalizeBuilderConfig.ts
robinmolen: I though it would be best to keep the `getGroupLabel` future proof. In case a new group is added in the backend, which isn't added in the formio-builder, that we can display the default title instead.

However, the formio-builder breaking because the input isn't supported, might indeed be a better setup

src/components/designer/ComponentsList.tsx
robinmolen: Yes, we can. The change is relatively small, and it does cleanup some additional logic (like the `expanded` state that is no-longer needed)

src/components/designer/normalizeBuilderConfig.ts
sergei-maertens: Yeah I prefer the explicit approach and having the type system help me with what's valid configuration and what not. There are constructs where you can get a union of string literals but can also pass a plain string, however that seems like overkill for this situation where we want to keep tight control anyway.

src/components/designer/ComponentsList.tsx
sergei-maertens: it might also enable us to collapse/expand multiple groups, something I've found annoying before in the legacy UI :-) but that's not a hard requirement, more like a possible pleasant side effect

src/components/designer/ComponentsList.tsx
robinmolen: check.. I've updated the setup so that the filtering is done inside the component that would render the groups. I don't think `useDeferredValue` will do a lot in this case, as we already have all the information in the component and the searching/filtering is extremely light/fast. (When testing the differences between using `useDeferredValue` and not in storybook, I haven't noticed any changes) I'll hold back implementing the `useDeferredValue`, until we notice any UI issues.

I've also removed `useMemo`, as there isn't any actual gains with the `useMemo` setup (i think the `components.filter` is performed the same amount of time either way). In case there are actual `components.filter` calls that would not have been executed with a `useMemo` setup, its probably still lighter to run a couple more `.filter` calls then the added state management of multiple `useMemo` instances.

src/components/designer/ComponentsList.tsx
robinmolen: Oh yeah, the collapsing/expanding is super simple. By using the same `name` for multiple `details` we can provide the "only one can be opened at a time" functionality, and a simple `open` prop lets us forcefully open a specific `details` element.

I haven't found a solution for forcing one `details` to be open at any time, but i think thats negligible. (Its already much better then the current implementation, which just breaks when you use the search input and clear it without adding a component)

src/registry/addressNL/index.ts
```diff
@@ -12,6 +13,10 @@ export default {
   preview: {
     panel: Preview,
   },
+  formDesigner: {
+    icon: 'home',
+    label: intl => intl.formatMessage(COMPONENT_TYPE_LABELS.addressNL),
```
sergei-maertens: does this need to be dynamic? Can't you set the type of `label` to `MesageDescriptor` and then invoke the `intl.formatMessage` in the generic code?

src/registry/types.ts
```diff
@@ -49,12 +49,25 @@ export interface Previews {
...
 
+export interface FormDesigner {
+  /**
+   * The icon used in the form designer.
+   */
+  icon: string;
+  /**
+   * The label used in the form designer.
+   */
+  label: (intl: IntlShape) => string;
```
sergei-maertens:
```suggestion
  /**
   * The label used in the form designer.
   */
  label: MessageDescriptor;
```
This allows you to call `intl.formatMesasge(formDesigner.label)` in the generic code, avoiding a layer of indirection. You then also don't need the `messages.ts` module and can inline the label for each component in the registry through `defineMessage({...}),`, which makes it properly self contained.

Removing a component is then still as easy as `rm -rf src/registry/<type>` without also having to touch files in other places (which you would have to do now with the `src/registry/messages.ts` file)

src/components/designer/ComponentsList/ComponentsList.tsx
```diff
@@ -0,0 +1,67 @@
...
+          description: 'Components list search input accessible label',
+          defaultMessage: 'Search component(s)',
+        })}
+      />
+
+      {groups
+        .sort((groupA, groupB) => groupA.weight - groupB.weight)
+        .map(group => (
+          <RegularComponentsGroup
+            key={`${id}-${group.name}`}
```
sergei-maertens: Why are you including the `useId()` value in the key?

I think this can just use the group name + group index as key as stable identity that won't have collisions. I'd even be tempted to go with the `index` alone because you order it explicitly already, and we don't really expect this to re-render at all since the configuration is static/hardcoded in the code.

src/components/designer/types.ts
```diff
@@ -0,0 +1,55 @@
+import type {AnyComponentSchema} from '@open-formulieren/types';
+
+import type {BuilderInfo} from '@/components/ComponentEditForm';
+
+/**
+ * The definition of a component, as it is available in the backend project.
+ */
+export type ComponentDefinition = {
+  builderInfo: Omit<BuilderInfo, 'schema'> & {schema: AnyComponentSchema};
+};
```
sergei-maertens: Can we instead update the `BuilderInfo` type and drop the dependency on formiojs entirely for this type?

Ultimately, the type here is what we want to support, and not the extended component schema from upstream formiojs.

src/components/designer/types.ts
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * Configuration for a custom preset component.
+ */
+export type PresetComponentDefinition = Pick<BuilderInfo, 'title' | 'schema'> & {
+  key: string;
+  icon: string;
+};
```
sergei-maertens, line 11: Odd to keep the too-vague/loose type for `schema` here, can't we use our `AnyComponentSchema` here too? And at this point when picking only two properties, I'd be tempted to just completely define the type first-party instead of deriving it from another type.


```suggestion
/**
 * Configuration for a custom preset component.
 */
export interface PresetComponentDefinition {
  title: React.ReactNode;
  schema: AnyComponentSchema;
  key: string;
  icon: string;
};
```

src/components/designer/types.ts
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * The supported component group names.
+ */
+export type GroupName = 'custom' | 'custom_layout' | 'custom_special';
```
sergei-maertens: afaik these groups don't have any special meaning, so if possible I'd like to see those properly cover their meaning:


```suggestion
/**
 * The supported component group names.
 */
export type GroupName = 'basic' | 'special' | 'layout';
```
so that we use [consistent nomenclature like in the types](https://github.com/open-formulieren/types/blob/e9cc803c0a102f61e1973a6bed88aae7e76ff9ce/src/components/index.ts#L46).

src/components/designer/types.ts
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * The builder group configuration as its available in the backend project.
```
sergei-maertens:
```suggestion
 * The builder group configuration as it's available in the backend project.
```

src/components/designer/types.ts
```diff
@@ -0,0 +1,55 @@
...
+export interface ComponentGroup {
+  /**
+   * Whether the group should be expanded by default. If no default group is set, the first
+   * group will be expanded.
+   */
+  default?: true;
+  /**
+   * The title of the group.
+   */
+  title: string;
```
sergei-maertens:
```suggestion
  /**
   * The title of the group.
   */
  title: React.ReactNode;
```
could/should also work here and allows you to pass `FormattedMessage` elements

src/components/designer/types.ts
```diff
@@ -0,0 +1,55 @@
...
+  /**
+   * The technical name of the group.
+   */
+  name: GroupName;
+  /**
+   * The weight of the group.
+   */
+  weight: number;
```
sergei-maertens:
```suggestion
  /**
   * The weight of the group.
   * 
   * Higher weight is sorted to the bottom.
   */
  weight: number;
```
do check that this is correct - I'd interpret it as high weight falls down, but I can never remember what positive/negative value in `sort` actually produces in terms of ordering.

src/components/designer/types.ts
```diff
@@ -0,0 +1,55 @@
...
+  /**
+   * The components that should be shown in the group. If a component is not listed here, it
+   * will not be shown.
+   *
+   * Components can be listed as a boolean value, in which case the component
+   * configuration is fetched from the available form components. If a component is
+   * listed as a custom configuration, it will be shown with the given title and icon.
```
sergei-maertens: this part is not up to date any longer, is it?

src/components/designer/types.ts
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * The builder group configuration as its available in the backend project.
+ */
+export interface ComponentGroup {
+  /**
+   * Whether the group should be expanded by default. If no default group is set, the first
+   * group will be expanded.
+   */
+  default?: true;
```
sergei-maertens: do we need this? It's weird that you can now pass multiple groups that all have `default: true`. I'd expect that the default behaviour of expanding the first group is sufficient?

src/components/designer/ComponentsList/ComponentsGroup.tsx
```diff
@@ -0,0 +1,201 @@
...
+    }))}
+  />
+);
+
+type GroupComponentDefinition = {
+  key: string;
+  label: string;
+  icon: string;
+  isDeprecated?: boolean;
+  // The schema can be either a regular component schema or a custom-defined preset component schema.
```
sergei-maertens: These custom defined preset component schemas are still expected to comply with the schema types we define. With our types, we want to avoid:

* information from being missing for a particular schema, which would lead to runtime errors
* additional configuration being specified that we don't support, which wouldn't meet expectations and can lead to long debug sessions to try to figure out why "it's not working"

src/components/designer/ComponentsList/ComponentsGroup.tsx
```diff
@@ -0,0 +1,201 @@
...
+  componentDefinitions,
+}) => {
+  const id = useId();
+  const searchQuery = query.toLowerCase();
+  const isSearching = searchQuery !== '';
+
+  const filteredComponents = !isSearching
+    ? componentDefinitions
+    : componentDefinitions.filter(componentDefinition => {
+        const searchTerms = [componentDefinition.key, componentDefinition.label.toLowerCase()];
```
sergei-maertens: key should also be lowercased, no?

src/components/designer/ComponentsList/ComponentsGroup.tsx
```diff
@@ -0,0 +1,201 @@
...
+        return searchTerms.some(term => term.includes(searchQuery));
+      });
+
+  if (filteredComponents.length === 0) return null;
```
sergei-maertens: should the group as a whole be hidden, or should we display a message "No matching components found." instead?

src/components/designer/ComponentsList/ComponentsGroup.tsx
```diff
@@ -0,0 +1,201 @@
...
+      data-testid={testId}
+    >
+      <summary className="card-header bg-default">
+        <span className="card-title">{title}</span>
+      </summary>
+
+      <div className="card-body p-2">
+        <ul className="list-unstyled mb-0">
+          {filteredComponents.map((component, index) => (
+            <li key={`${id}-${index}`}>
```
sergei-maertens:
```suggestion
            <li key={`${component.type}-${component.key}`}>
```
the same component key and type combination should not occur, right?

sergei-maertens (changes requested):

src/components/designer/ComponentsList/ComponentsGroup.tsx
robinmolen: I've opted to hide the group in its entirely, to keep it close to the current implementation and reduce the amount of visual clutter.

Edit: Discussed with Lidwien. For now we keep it as is, so we only show the component categories/groups that have search results. When the search query results in zero results, we will display a "No matching components found." message. This means we need to move the filtering to the higher level

src/components/designer/types.ts
robinmolen: This is the group config that's passed from the backend config, so i think this shouldn't become a ReactNode element.
Actually, this `title` is no longer needed or used, as the titles are defined using a `defineMessages` to control translations. I'll remove the `title` property.

src/components/designer/types.ts
robinmolen: Will add. Higher is indeed ordered after lower values.

By default, the sort function uses ascending order, https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort:
> The sort() method of [Array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array) instances sorts the elements of an array [in place](https://en.wikipedia.org/wiki/In-place_algorithm) and returns the reference to the same array, now sorted. The default sort order is ascending, built upon converting the elements into strings, then comparing their sequences of UTF-16 code unit values.

For displaying the groups we do use a custom comparison function, but it still uses the default `sort()` logic of `(a, b) => a - b`, resulting in the collection being sorted in ascending order.

src/components/designer/ComponentsList/ComponentsGroup.tsx
robinmolen: Yeah right that's right..

src/components/designer/ComponentsList/ComponentsGroup.tsx
robinmolen: Mhh, right. But we cannot guarantee that with the custom presets. Even if we would implement some validation in the backend to ensure this, i still think we should play it save with actual unique key values.

But maybe replacing the `id` with `component.type + component.key` would be better/less overkill.

src/components/designer/types.ts
robinmolen: I kept the loose type here to allow the presets to use the Formio definitions of components, instead of our OF definitions.

However, using the OF definitions might indeed be better. To indeed keep the schema's consistent, but also so we can give some guarantee that the custom presets will work as expected.

Updating it

src/components/designer/types.ts
robinmolen: Check, will do. I've tried that before, but that caused some tests to break. I'll fix those too.

There are some other places using the `formio` dependency. I'll check if we can re-write those and remove the dependency in its entirely

sergei-maertens: > There are some other places using the formio dependency. I'll check if we can re-write those and remove the dependency in its entirely

As long as it's directly related to this PR/issue. I don't want to bloat this PR further and aim for gradual phasing out, if possible through a separate PR for the last bits (depending on how much that is), but don't want to get too distracted from the main goal here for 4.0!

src/registry/addressNL/index.ts
robinmolen: I've done it like this so we don't have to run `intl.formatMessage` for every search query change (as it uses the translated label as search term). But somewhere along the way i seem to have changed this, as the labels are translated/transformed before the filtering...

I'll double check if we can now indeed simplify this

src/components/designer/types.ts
robinmolen: Yeah indeed, removing it

src/components/designer/ComponentsList/ComponentsList.tsx
robinmolen: Good point...

`index` or `group.name` should indeed be enough... I'll change it to `index`, as this is guaranteed to be unique

src/components/designer/ComponentsList/ComponentsList.stories.tsx
```diff
@@ -0,0 +1,1003 @@
...
+        schema: {
+          label: 'Website',
+          id: 'url',
+          autocomplete: 'url',
+          type: 'textfield',
+          key: 'url',
+        },
+      },
+    ],
+    components: {
```
sergei-maertens: I don't understand why this is necessary when we already have a registry with defaults for each component type :(

src/components/designer/ComponentsList/ComponentsList.stories.tsx
```diff
@@ -0,0 +1,1003 @@
...
+            key: 'customerProfile',
+            label: 'Customer profile',
+            digitalAddressTypes: ['email', 'phoneNumber'],
+            shouldUpdateCustomerData: false,
+          },
+        },
+      },
+    },
+  },
+} as Meta<typeof ComponentsList>;
```
sergei-maertens:
```suggestion
} satisfies Meta<typeof ComponentsList>;
```

src/components/designer/ComponentsList/ComponentsList.tsx
```diff
@@ -0,0 +1,136 @@
...
+    defaultMessage: 'Preset',
+  },
+});
+
+export interface ComponentsListProps {
+  /**
+   * Mapping of component types to their builder information. This information should be
+   * retrieved from the Formio registry in the backend.
+   */
+  components: Record<AnyComponentSchema['type'], ComponentDefinition>;
```
sergei-maertens: Isn't this information available in our *own* registry? I don't understand why the backend needs to provide this.

src/components/designer/ComponentsList/ComponentsList.tsx
```diff
@@ -0,0 +1,136 @@
...
+  const isSearching = deferredQuery !== '';
+
+  const normalizedComponentGroups: NormalizedComponentGroups[] = useMemo(() => {
+    // Sort the groups by weight and normalize the components in each group.
+    const normalizedGroups: NormalizedComponentGroups[] = groups
+      .sort((groupA, groupB) => groupA.weight - groupB.weight)
+      .map(group => ({
+        groupName: group.name,
+        components: normalizeComponentDefinition(components, group, intl),
+      }));
```
sergei-maertens: Seeing the weights specified in the stories like that, I'm actually questioning why we'd need to specify weight in the first place when we provide an array of groups. An array of groups already has ordering, we don't need the weights crutch to properly sort the groups - that's controlled by us when we pass the groups as prop already!

src/components/designer/ComponentsList/ComponentsList.tsx
```diff
@@ -0,0 +1,136 @@
...
+            testId={`component-group--${groupName}`}
+            title={<FormattedMessage {...COMPONENT_GROUP_LABELS[groupName]} />}
+            componentDefinitions={components}
+          />
+        ))
+      ) : (
+        <p className="text-muted" id={noSearchResultsMessageId}>
+          <FormattedMessage
+            description="Components list 'no results found' message"
+            defaultMessage="No matching components found"
```
sergei-maertens:
```suggestion
            defaultMessage="No matching components found."
```

src/components/designer/ComponentsList/normalizeComponentDefinition.ts
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * Normalize the component definitions for a given component group.
+ */
+export const normalizeComponentDefinition = (
+  components: Record<AnyComponentSchema['type'], ComponentDefinition>,
+  group: ComponentGroup,
+  intl: IntlShape
+) =>
+  group.components.map<NormalizedComponentDefinition>(componentType => {
+    const component = components[componentType];
```
sergei-maertens: I'd update `getRegistryEntry` from `@/registry` to allow a component type for lookup rather than a whole component (or support both). I'd also expect the registry to hold the builder info for each component type.

src/components/designer/ComponentsList/ComponentsList.stories.tsx
robinmolen: Well i thought it would be best to use the component definitions directly from the backend, as we know those are correct. Adding a definition within the formio-builder just adds another list of information that could get stale/faulty.

And i thought we then could still manage their config through the backend project, but that doesn't really make sense... (We can just... edit it here)

sergei-maertens: The formio-builder repo is supposed to be the source of truth, especially because that one actually type checks against our type definitions.

Everything in the backend only exists because of formio.js constructs that nobody wanted.

The defaults for the schema are already defined in the formio-builder since the very first versions, centralizing everything there with a clearly structured registry was always the intention.

src/components/designer/ComponentsList/normalizeComponentDefinition.ts
robinmolen: With moving the `builderInfo` to the registry, i've simplified the `formDesigner` config a bit. The `formDesigner` icon has been removed, as we can use the `builderInfo.icon`. I did however leave the translatable `MessageDescriptor` on the `formDesigner`, to prevent confusion about the shape of the `builderInfo`.

`BuilderInfo` is used in a couple more places, where a string value is expected instead of a React-intl object.

Again, because `builderInfo` is used in multiple places, the typescript definition has been left loose. So the weight, group and documentation are all optional. These values have NOT been entered in the component registries, as we don't need them for the form designer.

src/components/designer/ComponentsList/ComponentsList.tsx
robinmolen: Fair.. Removing it

src/components/designer/ComponentsList/ComponentsList.tsx
robinmolen: It is indeed no-longer needed. Using the newly added `BuilderInfo` on the registry, we can fetch the component schema through the registry

sergei-maertens (approved):
You'll have to rebase to deal with the merge conflicts of course :)

PR #277, opened 2026-04-15
JSON editor storybook mocking

### LINKED ISSUE #276: JSON editor mocking path resolve issues
The JSON editor mocking for storybook has some issues with path resolving. This results in vite internal errors when trying to run `storybook` or `test:storybook` locally.

The problem seems to be related to either running the commands from macOS, or when the path contains spaces.

robinmolen:
Closes #276

The previous mocking setup, using `new URL().pathname`, works differently across OS systems. The URLs are represented/encoded differently on macOS as on Linux systems, causing the path not to be resolved. Additionally, spaces or other special characters in the URL path can also cause issues with this setup.

By using `fileURLToPath` the URLs are 'correctly' decoded, and the mocking import can be resolved on a wider scale of systems/file structures.

### DIFF
--- a/.storybook/main.ts
+++ b/.storybook/main.ts
@@ -1,4 +1,5 @@
 import type {StorybookConfig} from '@storybook/react-vite';
+import {fileURLToPath} from 'node:url';
 
 const config: StorybookConfig = {
   core: {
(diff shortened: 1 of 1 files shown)


annashamray (approved):
It was a collateral damage of moving from cjs to esm 
Thanks for fixing!

PR #278, opened 2026-04-22
[OF#6106] Updated form type names according to backend

### LINKED ISSUE #6106: Adapt the form designer UI based on the new form type field
The `Form` now has a new field (`type`) which defines the type of the form, regular, appointment or a single page. With this change the admin UI has to be updated too (see the parent issue and the technical approach).

- [x] reset and hide login options
- [x] ensure only one step can be added/configured
- [x] prevent login requires on the single step
- [x] allow registration
- [x] allow logic
- [x] allow confirmation templates
- [x] disable prefill
- [x] disable/hide progress indicator display
- [x] disable submit options (always submit without overview)

vaszig:
Closes open-formulieren/open-forms#6106 partly

### DIFF
--- a/src/components/ComponentConfigurationSinglePage.stories.tsx
+++ b/src/components/ComponentConfigurationSinglePage.stories.tsx
@@ -131,7 +131,7 @@ export const Textfield: Story = {
       tooltip: 'An example for the tooltip',
       description: 'A description for the textfield component',
     },
-    formType: 'single_page',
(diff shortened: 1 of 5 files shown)


vaszig:
According to https://github.com/open-formulieren/open-forms/pull/6102 feedback.

sergei-maertens (approved):

PR #280, opened 2026-04-29
Feature/267 form designer component preview

### LINKED ISSUE #267: Form-designer - Form preview inputs

- [ ] textfield
- [ ] email
- [ ] date
- [ ] datetime
- [ ] time
- [ ] phoneNumber
- [ ] textarea
- [ ] number
- [ ] checkbox
- [ ] selectboxes
- [ ] select
- [ ] currency
- [ ] radio
- [ ] iban
- [ ] licenseplate
- [ ] postcode
- [ ] bsn
- [ ] signature
- [ ] cosign
- [ ] editgrid
- [ ] addressNL
- [ ] partners
- [ ] children
- [ ] customerProfile
- [ ] content
- [ ] columns
- [ ] fieldset
- [ ] softRequiredErrors
- [ ] coSign

### LINKED ISSUE #4517: form designer admin (not a blocker)

robinmolen:
Partly closes #267
Closes https://github.com/open-formulieren/open-forms/issues/4517

The basis for implementing the Formio-builder own form designer. This contains the component preview wrapper, and storybook implementations for all available Formio components. I've tried to keep the storybook setup as clean as possible, still its a lot to go through.

A bunch of components are included in the storybook tests, but aren't "done" yet. These components will be fixed/updated in separate PR's, to keep the overview a bit simpler. These "not yet finished" components are the `Fieldset`, `Editgrid`, `Columns`, `LeafletMap`, `Children`, `Content`, `SoftRequiredErrors` and `Currency`.
Most of these will simply display a white screen when checking their storybook page, this is expected. The `Currency` component will display a complete input, but will be replaced with something that closer resembles the current admin implementation and the public facing frontend.
These components can now mostly be ignored. These were included to provide a complete overview of all components available for this designer preview.

### DIFF
--- a/src/components/designer/Preview.stories.tsx
+++ b/src/components/designer/Preview.stories.tsx
@@ -1,0 +1,3574 @@
+import type {
+  AddressNLComponentSchema,
+  BsnComponentSchema,
+  CheckboxComponentSchema,
+  ChildrenComponentSchema,
+  ColumnsComponentSchema,
+  ContentComponentSchema,
+  CosignV2ComponentSchema,
+  CurrencyComponentSchema,
+  CustomerProfileComponentSchema,
+  DateComponentSchema,
+  DateTimeComponentSchema,
+  EditGridComponentSchema,
+  EmailComponentSchema,
+  FieldsetComponentSchema,
+  FileComponentSchema,
+  IbanComponentSchema,
+  LicensePlateComponentSchema,
+  NpFamilyMembersComponentSchema,
+  NumberComponentSchema,
+  PartnersComponentSchema,
+  PhoneNumberComponentSchema,
+  PostcodeComponentSchema,
+  RadioComponentSchema,
+  SelectComponentSchema,
+  SelectboxesComponentSchema,
+  SoftRequiredErrorsComponentSchema,
+  TextFieldComponentSchema,
+  TextareaComponentSchema,
+  TimeComponentSchema,
+} from '@open-formulieren/types';
+import type {Meta, StoryObj} from '@storybook/react-vite';
+import {expect, fireEvent, userEvent, waitFor, within} from 'storybook/test';
+
+import Preview from './Preview';
+
+export default {
+  title: 'Form designer/Preview',
+  component: Preview,
+  parameters: {
+    modal: {noModal: true},
+  },
+} as Meta<typeof Preview>;
+
+type Story = StoryObj<typeof Preview>;
+
+export const TextField: Story = {
+  args: {
+    components: [
+      {
+        type: 'textfield',
+        id: 'textfield',
+        key: 'textfieldPreview',
+        label: 'Textfield preview',
+        description: 'A preview of the textfield Formio component',
+        tooltip: 'A preview of the textfield Formio component',
+        defaultValue: 'Default value',
+        hidden: false,
+        placeholder: 'Sample placeholder',
+        showCharCount: true,
+      } satisfies TextFieldComponentSchema,
+      {
+        type: 'textfield',
+        id: 'textfieldHidden',
+        key: 'textfieldPreviewHidden',
+        label: 'Textfield preview hidden',
+        description: 'A preview of the textfield Formio component in hidden state',
+        tooltip: 'A preview of the textfield Formio component in hidden state',
+        defaultValue: 'Default value',
+        hidden: true,
+        placeholder: 'Sample placeholder',
+        showCharCount: true,
+      } satisfies TextFieldComponentSchema,
+    ],
+  },
+  play: async ({canvasElement, step}) => {
+    const canvas = within(canvasElement);
+
+    // Expect label, input and description to be shown
+    const regularInput = canvas.getByLabelText('Textfield preview');
+    expect(regularInput).toBeVisible();
+    expect(canvas.getByText('A preview of the textfield Formio component')).toBeVisible();
+
+    // Expect the label, input and description of the hidden component to be visible
+    const hiddenInput = canvas.getByLabelText('Textfield preview hidden');
+    expect(hiddenInput).toBeVisible();
+    expect(
+      canvas.getByText('A preview of the textfield Formio component in hidden state')
+    ).toBeVisible();
+
+    // Expect the wrapper of the hidden component to have a descriptive "is hidden" title
+    const hiddenInputWrapper = hiddenInput.closest('[data-testid="designerPreview"]');
+    expect(hiddenInputWrapper).toBeInTheDocument();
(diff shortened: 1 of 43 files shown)


src/components/designer/Preview.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+import './Preview.scss';
+import {DragContainer, DraggableComponent} from './drag-n-drop';
+
+export interface ComponentsPreviewProps {
+  components: AnyComponentSchema[];
+}
+
+export const ComponentsPreview: React.FC<ComponentsPreviewProps> = ({components}) => {
+  const initialValues = components.reduce<Record<string, any>>((carry, component) => {
```
SonnyBA: Can we make the value type (of the `Record`) a bit more specific here? In open forms we have a [`JSONValue`](https://github.com/open-formulieren/open-forms/blob/0d9e1dc06062c8a9e5c32630eaf49860c691f3ae/src/openforms/typing.py#L26) type but I'm not sure there is something similar in Typescript.

src/components/designer/Preview.stories.tsx
SonnyBA, line 1: Is it possible to split this into multiple files (per component) under a `Preview` folder?

src/components/designer/Preview.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+export const ComponentPreview: React.FC<ComponentPreviewProps> = ({component}) => {
+  const intl = useIntl();
+  const entry = getRegistryEntry(component);
+  const {
+    preview: {designer: PreviewComponent},
+  } = entry;
+  // @TODO Remove the undefined check when all components have a designer preview.
+  if (PreviewComponent === undefined || PreviewComponent === null) {
+    return null;
```
SonnyBA, line 72: We could log/print the components here that have no preview component so we don't have to rely on manually checking each component for the attribute 🤔 .

robinmolen: Good point.. In theory, each component should have a preview, as the form designer will otherwise break... I think that print statement will only be a temporary addition, but i can definitely include it 👍

For now the components that are missing here are the `fieldset`, `editgrid`, `columns`, `map`, `children`, `content`, and `soft required errors`. Im currently working on adding each one of these

src/components/designer/Preview.stories.tsx
robinmolen: Mhhh, yeah something like that also crossed my mind, but decided against it.

I placed them all in the `Preview.stories.tsx` because those stories are all displays of the `Preview` component for different arguments (i.e. which components should be displayed). This is similar to the `ComponentPreview.stories.tsx` setup.

Additionally, the individual component form designer previews are located in the component registries. So the individual 'stories' files would also not accompany anything. Im not sure if individual storybook file will improve the readability here

src/components/designer/Preview.tsx
robinmolen: Check, i'll take a look. `any` is indeed a bit nasty here..

src/components/designer/Preview.tsx
```diff
@@ -0,0 +1,102 @@
+import type {AnyComponentSchema} from '@open-formulieren/types';
```
SonnyBA:
```suggestion
import type {AnyComponentSchema, JSONValue} from '@open-formulieren/types';
```

src/components/designer/Preview.tsx
```diff
@@ -0,0 +1,102 @@
...
+import clsx from 'clsx';
+import {Formik} from 'formik';
+import {useIntl} from 'react-intl';
+
+import ErrorBoundary from '@/components/error/ErrorBoundary';
+import {getRegistryEntry} from '@/registry';
+import {hasOwnProperty} from '@/types';
+import type {JSONValue} from '@/types';
```
SonnyBA:
```suggestion
```

src/registry/types.ts
```diff
@@ -1,8 +1,9 @@
-import {AnyComponentSchema} from '@open-formulieren/types';
-import {IntlShape} from 'react-intl';
+import type {AnyComponentSchema} from '@open-formulieren/types';
```
SonnyBA:
```suggestion
import type {AnyComponentSchema, JSONValue} from '@open-formulieren/types';
```

src/registry/types.ts
```diff
@@ -1,8 +1,9 @@
...
+import type {IntlShape} from 'react-intl';
 import {z} from 'zod';
 
-import {BuilderContextType} from '@/context';
+import type {BuilderContextType} from '@/context';
+import type {JSONValue} from '@/types';
```
SonnyBA:
```suggestion
```

src/types/index.ts
```diff
@@ -1,5 +1,6 @@
 export * from './schemas';
 export * from './paths';
+export * from './json';
```
SonnyBA:
```suggestion
```

src/types/json.ts
```diff
@@ -0,0 +1,5 @@
+export type JSONPrimitive = string | number | boolean | null;
+export type JSONValue = JSONPrimitive | JSONValue[] | JSONObject;
```
SonnyBA: I saw the `types` repository [has the same type available](https://github.com/open-formulieren/types/blob/e9cc803c0a102f61e1973a6bed88aae7e76ff9ce/src/json.ts#L2) (which has the type definition as defined here).

src/components/designer/Preview.stories.tsx
```diff
@@ -0,0 +1,3574 @@
...
+    await step('Interaction with hidden email field', async () => {
+      // check that user can type into the fields
+      await userEvent.clear(hiddenInput);
+      await userEvent.type(hiddenInput, 'hello@example.com');
+      await expect(hiddenInput).toHaveDisplayValue('hello@example.com');
+    });
+  },
+};
+
+export const EmailWithVerification: Story = {
```
SonnyBA, line 444: I noticed the styling for the "verify" button was a bit off here:

Image

src/types/json.ts
robinmolen: Oh wow, i completely missed that... I was looking at the renderer project, and saw we defined the JSON types there directly, so though we didn't have them centralized.

src/components/designer/Preview.stories.tsx
robinmolen: The placement is indeed not correct, looking into it. This should be like the `email with verification` component preview (the preview used when editing a email component).

The color is correct, this is updated/corrected in the SDK to be conform that theme and color usage.

robinmolen: I've updated the styling, which fixes the alignment of the input and button

SonnyBA (approved):

PR #281, opened 2026-04-29
Form designer fieldset preview

### LINKED ISSUE #267: Form-designer - Form preview inputs

robinmolen:
Partly closes #267

Implementing the form designer preview component for `Fieldset` Formio components.

### DIFF
--- a/i18n/messages/en.json
+++ b/i18n/messages/en.json
@@ -694,6 +694,11 @@
     "description": "Tooltip for 'IsSensitiveData' builder field",
     "originalDefault": "The data entered in this component will be removed in accordance with the privacy settings."
   },
+  "JGqzCP": {
+    "defaultMessage": "Drag a component in the form and release the mouse button",
+    "description": "Fieldset designer preview content description",
+    "originalDefault": "Drag a component in the form and release the mouse button"
+  },
   "JKnHMC": {
     "defaultMessage": "Lastname",
     "description": "Label for partner's lastname",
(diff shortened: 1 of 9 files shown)


SonnyBA (commented):
I noticed the font size of the tooltip for fieldsets (24px) was very large compared to how the current admin looks (14px). Not sure if we can ignore that or that the tooltip styling needs to be adjusted.

robinmolen:
> I noticed the font size of the tooltip for fieldsets (24px) was very large compared to how the current admin looks (14px). Not sure if we can ignore that or that the tooltip styling needs to be adjusted.

Nice catch, i've seen that indeed as well in the admin, but didn't cross my mind when implementing. I think we should indeed keep that consistent. I've added a css `font-size` rule to the tooltip component

sergei-maertens (approved):

PR #284, opened 2026-05-07
Add builder configuration fields for new document type configuration

### LINKED ISSUE #283: Support specifying catalogue + IOT.omschrijving for file component registration options
To be able to migrate away from the document type URLs, the catalogue identifier and documenttype identifiers should be specified so that an (automated) migration script can convert the existing configuration into the new one.

For the time being, this means that we keep the registration options for the file component still in the registration tab.

We can release this as a new feature and it looks like the stable/3.5.x backend can upgrade to latest formio-builder without major issues - from the changelog it only appears that the customer profile email recipient option can cause confusion, but this can be addressed with documentation.

This will require an update to the types repository too.

sergei-maertens:
Closes #283 - depends on https://github.com/open-formulieren/types/pull/93

To be able to migrate and visualize the result from migration scripts, the builder must allow configuring the document type reference by catalogue identifiers and document type reference.

This is a minimal approach without much useful validation, as it needs to be applied to 3.5.x stable for people to be able to migrate in preparation for the 4.0 upgrade.

### DIFF
--- a/src/registry/file/edit-validation.ts
+++ b/src/registry/file/edit-validation.ts
@@ -68,11 +68,76 @@ const buildFileMaxSizeSchema = (intl: IntlShape, builderContext: BuilderContextT
     });
 };
 
+const buildRegistrationSchema = (intl: IntlShape) =>
+  z.object({
+    documentType: z
+      .object({
+        catalogue: z.object({
+          domain: z.string().optional(),
+          rsin: z.string().optional(),
+        }),
+        description: z.string().optional(),
+      })
+      .optional()
+      .superRefine((documentType, ctx) => {
+        const hasAnyProperty =
+          !!documentType?.description ||
+          !!documentType?.catalogue.domain ||
+          !!documentType?.catalogue.rsin;
+        if (hasAnyProperty) {
+          if (!documentType.description) {
+            ctx.addIssue({
+              code: z.ZodIssueCode.custom,
+              message: intl.formatMessage({
+                description:
+                  'File registration validation error for missing document type description',
+                defaultMessage:
+                  'You must specify a document type description when a catalogue is configured.',
+              }),
+              path: ['description'],
+            });
+          }
+          if (!documentType?.catalogue.domain) {
+            ctx.addIssue({
+              code: z.ZodIssueCode.custom,
+              message: intl.formatMessage({
+                description:
+                  'File registration validation error for missing catalogue domain or RSIN',
+                defaultMessage: 'You must specify both domain and RSIN.',
+              }),
+              path: ['catalogue', 'domain'],
+            });
+          }
+          if (!documentType?.catalogue.rsin) {
+            ctx.addIssue({
+              code: z.ZodIssueCode.custom,
+              message: intl.formatMessage({
+                description:
+                  'File registration validation error for missing catalogue rsin or RSIN',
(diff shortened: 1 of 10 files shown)


src/registry/file/edit-validation.ts
```diff
@@ -68,11 +68,76 @@ const buildFileMaxSizeSchema = (intl: IntlShape, builderContext: BuilderContextT
...
+            ctx.addIssue({
+              code: z.ZodIssueCode.custom,
+              message: intl.formatMessage({
+                description:
+                  'File registration validation error for missing catalogue rsin or RSIN',
+                defaultMessage: 'You must specify both domain and RSIN.',
+              }),
+              path: ['catalogue', 'rsin'],
+            });
+          } else if (!/[0-9]{9}/.test(documentType.catalogue.rsin)) {
```
annashamray:
```suggestion
          } else if (!/^[0-9]{9}$/.test(documentType.catalogue.rsin)) {
```

Should it be anchored?

src/registry/file/registration-tab.tsx
```diff
@@ -75,6 +78,109 @@ const DocumentTypeSelect: React.FC<{}> = () => {
   );
 };
 
+/**
+ * 'Modern' document type configuration.
+ *
+ * @todo Validate length of RSIN to be 9 characters and alpha-numeric.
```
annashamray: I think you already added it inthe zod validation. Or do you mean backend part?

src/registry/file/registration-tab.tsx
```diff
@@ -75,6 +78,109 @@ const DocumentTypeSelect: React.FC<{}> = () => {
...
+          />
+        }
+        tooltip={
+          <FormattedMessage
+            description="Tooltip for 'registration.documentType.catalogue.rsin' builder field"
+            defaultMessage="The 'rsin' attribute for the Catalogus resource in the Catalogi API."
+          />
+        }
+      />
+      <TextField
```
annashamray, line 144: The deprecated `informatieobjecttype` is a dropdown. Maybe we can also have a dropdown here? It can be easier for beheers

src/registry/file/registration-tab.tsx
sergei-maertens: I didn't clean up the TODO :-)

src/registry/file/registration-tab.tsx
sergei-maertens: Yeah but that invites a lot more complexity, and we have the components for that already in the backend. Ultimately this whole registration tab should be gone for all components, so I plan to tackle that at that point, not in a "fix" that needs to be backported to 3.5

Similarly in the django admin for the API group you now also configure the individual fields rather than with dropdowns.

src/registry/file/edit-validation.ts
sergei-maertens: Added!

src/registry/file/registration-tab.tsx
sergei-maertens: TODO removed

src/registry/file/registration-tab.tsx
annashamray: Ok! Thanks for the explanation

annashamray (approved):

PR #287, opened 2026-05-11
Implement content component designer preview

### LINKED ISSUE #267: Form-designer - Form preview inputs

robinmolen:
Partly closes #267

Implementing the `content` component form designer preview component

### DIFF
--- a/src/registry/content/designer-preview.scss
+++ b/src/registry/content/designer-preview.scss
@@ -1,0 +1,4 @@
+.offb-designer-preview.offb-designer-preview--component-type-content {
+  padding-inline: 10px;
+  border: dashed 2px #dddddd;
+}
--- a/src/components/designer/Preview.stories.tsx
+++ b/src/components/designer/Preview.stories.tsx
@@ -3033,7 +3033,6 @@ export const LicensePlateMultiple: Story = {
   },
 };
 
-// @TODO implement preview component and play
 export const Content: Story = {
   name: 'Content',
   args: {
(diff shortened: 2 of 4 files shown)


src/registry/content/designer-preview.scss
```diff
@@ -0,0 +1,5 @@
+.offb-designer-preview.offb-designer-preview--component-type-content {
+  margin-block-end: 0;
```
SonnyBA: Restoring the margin here to me makes it a bit more clear that there are two separate components (a regular and a hidden component) and I think this is also consistent with the other components?

SonnyBA (approved):
Looks good to me, one comment regarding the margin of the component but is not a blocker for me.

src/registry/content/designer-preview.scss
robinmolen: Yeah, i opted to remove the margin, to keep it as close as possible to the current admin implementation. I just checked if the `content` components are rendered without margin in the frontend (that the lack of margin in the builder would be an accurate representation of the user UI), but thats also not the case.

So lets do it with margin 👍

PR #289, opened 2026-05-11
Implement map designer preview

### LINKED ISSUE #267: Form-designer - Form preview inputs

robinmolen:
Partly closes #267

Implementing the form designer preview component for the `map` component

For now we re-use the regular map component preview component. The map interactions might interfere with the drag-n-drop behaviour, but we need the drag-n-drop to know this for sure.

If we do end up needing a custom designer preview for the map component, we can add that later.

### DIFF
--- a/src/components/designer/Preview.stories.tsx
+++ b/src/components/designer/Preview.stories.tsx
@@ -3617,7 +3617,6 @@ export const Signature: Story = {
   },
 };
 
-// @TODO implement preview component and play
 export const LeafletMap: Story = {
   name: 'Map',
   args: {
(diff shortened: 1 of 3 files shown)


src/components/designer/Preview.stories.tsx
```diff
@@ -3348,19 +3347,41 @@ export const LeafletMap: Story = {
         id: 'wekruya',
         type: 'map',
         key: 'map',
-        label: 'A map preview',
+        label: 'Map preview',
+        description: 'A preview of the map Formio component',
+        interactions: {marker: true, polyline: false, polygon: false},
```
SonnyBA: Minor detail but I think enabling all interactions here gives the designer an impression of how it would look whenever all features are enabled.

SonnyBA (approved):

src/components/designer/Preview.stories.tsx
robinmolen: Fair point. With that, i've also added tooltips and fixed a small z-index issue

PR #294, opened 2026-06-02
Implementing core form designer drag-and-drop behavior

### LINKED ISSUE #268: Form-designer - Basic D&D
Implementing the basic drag n drop behaviour of dragging a component from the component list to the preview

robinmolen:
Closes #268

This implements drag-and-drop behavior into the form designer. This is limited to visual drag-and-drop behavior. The actual updating for component configurations (like when moving a component from one layout component to another), will be implemented in #269.

So, what is done:

* Added dnd-kit and immer dependencies
* Setup core drag-and-drop behavior (moving components inside a dropzone)
* Using the ComponentsList component to creating new components
* Some initial setup for drag-and-drop behavior inside layout components (a lot of dropzone stuff will be updated/cleaned up in #269)
* Added a simple FormBuilder component, which will replace the Formio FormBuilder component at some point


src/registry/columns/designer-preview.tsx
```diff
@@ -15,37 +12,31 @@ const Preview: React.FC<ComponentPreviewProps<ColumnsComponentSchema>> = ({compo
...
-      </ErrorBoundary>
+      <div className="row" data-testid={`columns-${key}`}>
+        {columns.map((column, index) => (
+          <div
+            className={clsx('col', `col-${column.size}`)}
+            data-testid={`${key}-column-${index}`}
+            key={index}
+          >
+            <ComponentsPreview
+              dropzoneId={`${key}-${index}-dropzone`}
```
robinmolen: In #269 im adding some dropzone utils, that construct the `dropzoneId`'s. So this will be replaced with a `getColumnDropzoneId` function call

src/registry/editgrid/designer-preview.tsx
```diff
@@ -23,24 +19,7 @@ const Preview: React.FC<ComponentPreviewProps<EditGridComponentSchema>> = ({comp
...
-            <ContentPlaceholder variant="designer">
-              <FormattedMessage
-                description="Editgrid designer preview content description"
-                defaultMessage="Drag a component in the form and release the mouse button"
-              />
-            </ContentPlaceholder>
-          )}
-        </DropZone>
-      </ErrorBoundary>
+      <ComponentsPreview dropzoneId={`${key}-dropzone`} components={components} />
```
robinmolen: Same as above, the `dropzoneId` will be constructed using dropzone utils in #269

src/registry/fieldset/designer-preview.tsx
```diff
@@ -1,38 +1,16 @@
...
-            <ContentPlaceholder variant="designer">
-              <FormattedMessage
-                description="Fieldset designer preview content description"
-                defaultMessage="Drag a component in the form and release the mouse button"
-              />
-            </ContentPlaceholder>
-          )}
-        </DropZone>
-      </ErrorBoundary>
+      <ComponentsPreview dropzoneId={`${key}-dropzone`} components={components} />
```
robinmolen: Again, `dropzoneId` will be updated in #269

src/components/designer/ComponentsList/ComponentsGroup.tsx
```diff
@@ -1,6 +1,7 @@
 import clsx from 'clsx';
 import {FormattedMessage} from 'react-intl';
 
+import {DraggableMenuItem} from '../dragDrop';
 import type {NormalizedComponentDefinition} from '../types';
```
sergei-maertens: Please use absolute imports instead of relative, when I see `../` paths in a PR I've no idea where actually in the file structure this code is located and I need to calculate it. Using `@/components/designer/dragDrop` is much more explicit.

Relative imports in the same directory are allowed: `./foo` as that's easy to parse.

src/components/designer/ComponentsList/ComponentsList.tsx
```diff
@@ -1,12 +1,9 @@
...
 
-import type {
-  ComponentGroup,
-  GroupName,
-  NormalizedComponentGroups,
-  PresetComponentDefinition,
-} from '../types';
+import {DesignerContext} from '@/context';
+
+import type {GroupName, NormalizedComponentGroups} from '../types';
```
sergei-maertens: Can you also make this import absolute?

src/components/designer/ComponentsList/index.ts
```diff
@@ -0,0 +1 @@
+export {default as ComponentsList} from './ComponentsList';
```
sergei-maertens: This can just remain the default export, no? The component name matches the folder:

```tsx
import ComponentsList from '@/components/designer/ComponentsList';
```
is logical

src/components/designer/dragDrop/DropZone.scss
```diff
@@ -1,14 +1,22 @@
 .offb-drop-zone {
   height: 100%;
-  padding: 10px;
-  border: dotted 2px #e8e8e8;
+  display: flex;
+  flex-direction: column;
+  gap: 1rem;
```
sergei-maertens: we don't use rems anywhere else for spacing values, please stick to px.

src/components/designer/dragDrop/DropZone.scss
```diff
@@ -1,14 +1,22 @@
...
   &:hover {
     border-color: #ccc;
   }
 
-  .offb-designer-preview,
-  .offb-content-placeholder {
-    margin-block-end: 1rem;
+  .offb-designer-preview {
+    .form-group {
+      // The margin-bottom for the preview items is managed bt the drop-zone component.
```
sergei-maertens:
```suggestion
      // The margin-bottom for the preview items is managed by the drop-zone component.
```

src/components/designer/dragDrop/DropZone.scss
```diff
@@ -1,14 +1,22 @@
...
+      margin-bottom: 0;
```
sergei-maertens: can't we use the logical property `margin-block-end`?

src/components/designer/dragDrop/utils/dragTarget.ts
sergei-maertens, line 1: this seems very related to drag and drop - why isn't it in the `dragDrop` folder?

src/components/designer/dragDrop/utils/dropzone.ts
sergei-maertens, line 1: Same here about the location of this module, shouldn't this fit better in `dragDrop`?

src/components/ContentPlaceholder.tsx
sergei-maertens, line 1: `ComponentName/index.tsx` might as well just be `ComponentName.tsx`, no?

There's precedent with a bunch of other components that just have their component file.

src/components/designer/utils/fields.ts
```diff
@@ -0,0 +1,181 @@
+import type {AnyComponentSchema} from '@open-formulieren/types';
+import type {Draft} from 'immer';
+
+import {getRegistryEntry} from '@/registry';
+
+import type {ComponentPlaceholder} from '../types';
+import {COMPONENT_PLACEHOLDER_TYPE} from '../types';
```
sergei-maertens: please make these absolute imports

src/components/designer/utils/fields.ts
```diff
@@ -0,0 +1,181 @@
...
+      for (let index = 0; field.columns.length > index; index++) {
+        similarKeys.push(
+          ...findComponentKeysStartingWith(startsWith, field.columns[index].components)
+        );
+      }
+    }
+  }
+
+  return similarKeys;
+};
```
sergei-maertens: For the Formio processing, I'd prefer a similar implementation like in the renderer: https://github.com/open-formulieren/formio-renderer/blob/e62f3c7b6a454fd9d0bc70fcb8c49d5feba423ff/src/formio.ts#L155

Maybe in due time we can merge that then and use the renderer as dependency of the builder.

I find the generator approach easier to reason about.

src/components/designer/utils/fields.ts
```diff
@@ -0,0 +1,181 @@
+import type {AnyComponentSchema} from '@open-formulieren/types';
+import type {Draft} from 'immer';
+
+import {getRegistryEntry} from '@/registry';
+
+import type {ComponentPlaceholder} from '../types';
+import {COMPONENT_PLACEHOLDER_TYPE} from '../types';
+
+type Field = AnyComponentSchema | ComponentPlaceholder;
```
sergei-maertens: I know we use field/component interchangeably often, but I'm not sure that introducing a `Field` type here makes things more clear. Isn't "component definition" a better name?

src/components/designer/types.ts
```diff
@@ -1,5 +1,15 @@
 import type {AnyComponentSchema} from '@open-formulieren/types';
 
+export const COMPONENT_PLACEHOLDER_TYPE = 'componentPlaceholder';
+/**
```
sergei-maertens, line 5:
```suggestion
export const COMPONENT_PLACEHOLDER_TYPE = 'componentPlaceholder';

/**
```

src/components/designer/utils/fields.ts
```diff
@@ -0,0 +1,181 @@
...
+      similarKeys.push(...findComponentKeysStartingWith(startsWith, field.components));
+    }
+
+    if (field.type === 'columns') {
+      for (let index = 0; field.columns.length > index; index++) {
+        similarKeys.push(
+          ...findComponentKeysStartingWith(startsWith, field.columns[index].components)
+        );
+      }
+    }
```
sergei-maertens: you don't need the index, right? You can just loop over `for (const column of field.columns)`, no?

src/components/designer/utils/fields.ts
```diff
@@ -0,0 +1,181 @@
...
+/**
+ * Create a unique component key for the given component type.
+ *
+ * It looks through the existing fields and searches for component keys that start with
+ * the component type. If there are components with similar keys, it will append a number
+ * to the end of the key to make it unique.
+ * This is similar to the current Formio form builder implementation.
+ */
+const createComponentKey = (componentType: AnyComponentSchema['type'], fields: Field[]): string => {
+  const componentsWithSimilarKeys = findComponentKeysStartingWith(componentType, fields);
```
sergei-maertens: keep in mind that the backend uses *translated* component type names, so `textfield` -> `tekstveld`.

This should be kept, so we'll need some translation mappings and take into account the active language.

src/components/designer/utils/fields.ts
```diff
@@ -0,0 +1,181 @@
...
+export const createComponent = (
+  componentType: AnyComponentSchema['type'],
+  fields: Field[]
+): AnyComponentSchema => {
+  const {edit, builderInfo} = getRegistryEntry(componentType);
+
+  // Define component with their editor default values, and some generic defaults.
+  return {
+    ...edit.defaultValues,
+    id: self.crypto.randomUUID(),
```
sergei-maertens:
```suggestion
    id: window.crypto.randomUUID(),
```
I'm not sure why you wouln't use `window`?

src/components/designer/utils/fields.ts
```diff
@@ -0,0 +1,181 @@
...
+    type: componentType,
+    key: createComponentKey(componentType, fields),
+    ...(edit.defaultValues.hasOwnProperty('label') ? {label: builderInfo.title} : {}),
+  } as AnyComponentSchema;
```
sergei-maertens:
```suggestion
  } satisfies AnyComponentSchema;
```
but the types for this function should probably be generic. Right now, your type definition allows for a different component type to be returned than what you put in via `componentType`.

```ts
function createComponent<S extends AnyComponentSchema>(componentType: S['type']): S {
  // ... impl.
}
```

src/components/designer/utils/fields.ts
```diff
@@ -0,0 +1,181 @@
...
+    if (field.type === 'columns') {
+      for (let index = 0; field.columns.length > index; index++) {
+        const column = field.columns[index];
+        if (`${field.key}.${index}` === key) return column.components;
+
+        const found = findDropzoneFieldsByKey(column.components, key);
+        if (found) return found;
+      }
+    }
+  }
```
sergei-maertens: this would probably benefit from the earlier linked `iterComponents` generator so that the formio definition processing is properly abstracted and not duplicated. Now formio concepts and component types (that *should* ideally be managed in the registry!) are mixed with drag-and-drop related code, creating spaghetti that will become hard to maintain over time.

I'd even be tempted to explore putting the `getChildComponents` in the registry here in the builder, so that we don't hardcode *anything* w/r to `component.type` in generic code.

src/components/designer/utils/fields.ts
```diff
@@ -0,0 +1,181 @@
...
+
+/**
+ * Create a unique component key for the given component type.
+ *
+ * It looks through the existing fields and searches for component keys that start with
+ * the component type. If there are components with similar keys, it will append a number
+ * to the end of the key to make it unique.
+ * This is similar to the current Formio form builder implementation.
+ */
+const createComponentKey = (componentType: AnyComponentSchema['type'], fields: Field[]): string => {
```
sergei-maertens: I suspect you'll only be calling this on the current form definition being edited, but our backend code now has a mechanism to merge all the component definitions from all form steps together to guarantee unique key generation _for the whole form_.

Did you take that into account or not?

src/components/designer/utils/fields.ts
```diff
@@ -0,0 +1,181 @@
...
+ * Recursively remove the placeholder from the fields.
+ */
+const removePlaceholderFromFields = (fields: Field[]): boolean => {
+  const index = fields.findIndex(field => field.type === COMPONENT_PLACEHOLDER_TYPE);
+
+  // If the placeholder is found, remove it.
+  if (index >= 0) {
+    fields.splice(index, 1);
+    return true;
+  }
```
sergei-maertens: is there only ever going to be at most one? I'd expect a more robust `fields.filter(...)` to work equally well and the flag if something is removed then becomes `return result.length < fields.length;`.

sergei-maertens: Looks like the return value is not used though?

src/components/designer/utils/fields.ts
```diff
@@ -0,0 +1,181 @@
...
+const removeComponentFromFields = (fields: Field[], key: string): boolean => {
+  const index = fields.findIndex(
+    field => field.type !== COMPONENT_PLACEHOLDER_TYPE && field.key === key
+  );
+
+  // If the placeholder is found, remove it.
+  if (index >= 0) {
+    fields.splice(index, 1);
+    return true;
+  }
```
sergei-maertens: I'd also opt for `.filter` here I guess

src/components/designer/FormDesigner.stories.tsx
```diff
@@ -0,0 +1,127 @@
...
+    const canvas = within(canvasElement);
+    const basicComponentsList = canvas.getByTestId('component-group--basic');
+    const textareaDraggableItem = await within(basicComponentsList).findByRole('button', {
+      name: 'Textarea',
+    });
+    const dropzone = canvas.getByTestId('main-dropzone');
+
+    // As there are no components, a "how to add components" message should be displayed.
+    expect(
+      within(dropzone).getByText('Drag a component in the form and release the mouse button')
```
sergei-maertens:
```suggestion
      within(dropzone).getByText('Drag a component in the form and release the mouse button.')
```

src/components/designer/FormDesigner.stories.tsx
```diff
@@ -0,0 +1,127 @@
...
+    ).toBeVisible();
+
+    await dragTo(textareaDraggableItem, dropzone);
+
+    // The textarea component should be added to the dropzone
+    expect(within(dropzone).getByTestId('input-textarea'));
+    // As the dropzone isn't empty anymore, the instructions message should be gone.
+    expect(
+      within(dropzone).queryByText('Drag a component in the form and release the mouse button')
```
sergei-maertens:
```suggestion
      within(dropzone).queryByText('Drag a component in the form and release the mouse button.')
```

src/components/designer/Preview.tsx
```diff
@@ -1,58 +1,81 @@
...
-      : defaultValue;
+  return (
+    <ErrorBoundary>
+      <DropZone id={dropzoneId}>
+        {!hasComponents && (
+          <ContentPlaceholder variant="designer">
+            {contentPlaceholder || (
+              <FormattedMessage
+                description="Components preview empty content description"
+                defaultMessage="Drag a component in the form and release the mouse button"
```
sergei-maertens:
```suggestion
                defaultMessage="Drag a component in the form and release the mouse button."
```

src/components/designer/Preview.tsx
```diff
@@ -1,58 +1,81 @@
...
+        {components.map((component, index) =>
+          component.type === COMPONENT_PLACEHOLDER_TYPE ? (
+            <ComponentPlaceholderPreview
+              key="placeholder"
+              componentType={component.componentType}
+            />
+          ) : (
+            <SortableComponent
+              key={component.key}
+              id={component.key}
```
sergei-maertens:
```suggestion
              id={component.id}
```
might be a little bit more robust? I can already picture people editing the JSON by hand and setting up duplicated keys for child components...

src/components/designer/Preview.tsx
sergei-maertens, line 1: So what happens now that there's no Formik container anymore? I'd expect the preview forms to emit errors/warnings/crash, but looking at the stories that doesn't seem to be the case?

src/components/FormBuilder.tsx
```diff
@@ -0,0 +1,60 @@
+import type {AnyComponentSchema, JSONValue} from '@open-formulieren/types';
+import {Formik} from 'formik';
+
+import type {ComponentGroup, PresetComponentDefinition} from '@/components/designer/types';
+import {DesignerContext} from '@/context';
+import {getRegistryEntry} from '@/registry';
+import {hasOwnProperty} from '@/types';
+
+import FormDesigner from './designer/FormDesigner';
```
sergei-maertens: I'm wondering if we should go with `FormioDefinitionDesigner` or `FormioConfigurationDesigner`.

"Form designer" is already an overloaded concept referring to either the person creating forms in OF, or the whole "steps and fields" section in the backend.

src/components/FormBuilder.tsx
```diff
@@ -0,0 +1,60 @@
...
+      ? componentDefaultValue ?? []
+      : componentDefaultValue ?? defaultValue;
+
+    carry[key] = previewDefaultValue;
+    return carry;
+  }, {});
+
+  return (
+    <DesignerContext.Provider value={{groups, presets}}>
+      <Formik
```
sergei-maertens, line 50: Ah check, it moved here :-)

src/components/FormBuilder.tsx
```diff
@@ -0,0 +1,60 @@
...
+   */
+  groups: ComponentGroup[];
+  /**
+   * Custom-defined component presets that are shown in the form builder components list.
+   */
+  presets: PresetComponentDefinition[];
+}
+
+const FormBuilder: React.FC<FormBuilderProps> = ({components, presets, groups}) => {
+  const initialValues = components.reduce<Record<string, JSONValue | {}>>((carry, component) => {
```
sergei-maertens, line 27: I'm kinda tempted to just bite the bullet and start depending on the formio-renderer, because you can then call [`extractInitialValues`](https://github.com/open-formulieren/formio-renderer/blob/e62f3c7b6a454fd9d0bc70fcb8c49d5feba423ff/src/values.ts#L19) and have a proper working solution that's also more type-safe :)

src/registry/columns/designer-preview.tsx
```diff
@@ -15,37 +13,31 @@ const Preview: React.FC<ComponentPreviewProps<ColumnsComponentSchema>> = ({compo
...
+            <ComponentsPreview
+              dropzoneId={getColumnDropzoneId(component, index)}
+              components={column.components}
+              contentPlaceholder={
+                <>
+                  +
+                  <div className="sr-only">
+                    <FormattedMessage
+                      description="Column designer preview content description"
+                      defaultMessage="Drag a component in the form and release the mouse button"
```
sergei-maertens:
```suggestion
                      defaultMessage="Drag a component in the form and release the mouse button."
```
if the text is supposed to be the same as for fieldset, editgrid and the root drop zone, please use the same combination of `description` and `defaultMessage` so we only need to translate it once.

And, if that's the case, this can/should be a small component itself too instead of repeating that implementation and keeping it in sync.

src/components/designer/ComponentsList/ComponentsGroup.tsx
robinmolen: Im pretty sure this was prettier or our formatter... I'll double check

src/components/designer/Preview.tsx
robinmolen: Yeah moving it to a separate higher level component seemed better. So we can handle communications to the backend in isolation from all drag-and-drop and component rendering stuff

src/components/designer/utils/fields.ts
robinmolen: Yeah i just came across this when working on the floating actions. Currently we pass all components in a form to the Formio FormBuilder through `openForms.componentNamespace` builder options. But if there is a proper setup already in the backend, then that should be used instead..

I'll check it out

sergei-maertens: I wouldn't call it "a proper mechanism". We just need some kind of construct so that you can provide the *other* component definitions to the top-level component so that the namespace can be fully populated. Our generic comopnent definition processing (either from formio-builder or formio-renderer) can also process that list of formio configurations.

src/components/designer/dragDrop/utils/dragTarget.ts
robinmolen: I placed it in the designer level as it is solely used by the `FormDesigner`, but it indeed makes more sense in the `dragDrop` folder.

I'll move all these utils into the `dragDrop` folder, as each handles functionality needed for the drag and drop feature.

src/components/ContentPlaceholder.tsx
robinmolen: Do you mean just renaming that `index.tsx` file, or moving all `ContentPlaceholder` files to the `components` folder?

I'm fine with both

sergei-maertens: The latter :-)

src/components/designer/utils/fields.ts
robinmolen: Yes, there should at most only be one placeholder. The placeholder is used when dragging a new element onto the dropzones. When releasing the item, the placeholder gets replaced with an actual component instance.

So, if there is a placeholder in the components collection, that means that the user is actively adding a new component to the form designer.

Making this 'more robust' will also have some performance impact, as we no-longer will perform an early return. This will become more apparent when we introduce nested dropzones, where we have to iterate and create new arrays with every level.

The performance impact will probably be negligible, but adding any impact when we don't need it seems wrong

src/registry/columns/designer-preview.tsx
robinmolen: Check. I've replaced the `ComponentsPreview` custom 'empty' messages, with a `hideEmptyMessage` prop that makes the message visually hidden (screen-reader only). This ensures that the messages stay the same

src/components/designer/utils/fields.ts
robinmolen: Yeah sounds good!

I don't think we can re-use the renderer `iterComponents` (or at least, not the current setup), as it skips over editgrid components. But thats something for later :)

src/components/designer/utils/fields.ts
robinmolen: Yes agreed, I've renamed the types, file and usage

src/components/designer/utils/fields.ts
robinmolen: Check! I've removed this function for now, as it gave some issues with the `iterComponents` and `getChildComponents` (dealing with editgrid/fieldset vs columns, nested children, etc.). The idea of `getChildComponents` sounds nice, so i've integrated that into the `iterComponents`.

I took a look if we can also move the `iterComponents` to the registry, but thats a bit more difficult. The components collection here could also contain the placeholder, which isn't part of `AnyComponentSchema`. This causes issues when trying to make the input of `iterComponents` dynamic:

```
 TS2322: Type `AnyComponentSchema` is not assignable to type `S`
 `AnyComponentSchema` is assignable to the constraint of type `S`, but `S` could be instantiated with a different subtype of constraint `AnyComponentSchema`
 ```
 I suspect this is due to the layout children being typed to `AnyComponentSchema[]`, which is not assignable to type `S`.

I'll add the `findDropzoneFieldsByKey` function to the dnd layout implementation, as it is needed for that part.

File: src/components/designer/utils/fields.ts (hunk 17 above)
Comment (robinmolen) [reply src/components/designer/utils/fields.ts]: I've been trying for a while, but `satisfies` is not gonna happen. I think it is because the union created by `WithMultiple`, but im not 100% sure. I've tried a few things to get this to work, including making the `getRegistryEntry` more explicit:
```
export const getRegistryEntry = <S extends AnyComponentSchema>(
  componentType: S['type']
): RegistryEntry<S> => {
  const entry = REGISTRY[componentType];
  return entry;
};
```

Im gonna call it quits on this one. If you want, you can take a crack.


Last bit:
Changing the `EditFormDefinition` defaultValues did solve some of the `withMultiple` issues, but issues about component definitions not being assignable to `AnyComponentSchema` remain:
```
export type DistributiveOmit<T, K extends keyof any> = T extends any ? Omit<T, K> : never;

export interface EditFormDefinition<S extends AnyComponentSchema>
  extends React.FC<EditFormProps<S>> {
  defaultValues: DistributiveOmit<S, 'id' | 'type'>;
}
```

Comment (sergei-maertens) [reply src/components/designer/utils/fields.ts]: yeah that registry entry is crazy TS gymnastics and hard to solve.

It looks to me that the issue in your case is that TS can't narrow the discriminated union with the generic type because it doesn't know which component it actually is at runtime :thinking: 

I'll see if I can take a look, but these are definitely the cases where you run into type system limitations.

File: src/components/FormBuilder.tsx (hunk 28 above)
Comment (robinmolen) [reply src/components/FormBuilder.tsx:27]: Its a bit overkill for just the initial values, right? It also requires some real changes to the registry, to be able to supply the `getInitialValues` method (i've taken a quick look if we can re-use the formio-renderer registry, but got hit by a lot of unresolved paths)

I'll add it as task to https://github.com/open-formulieren/formio-builder/issues/271

File: src/components/designer/utils/fields.ts (hunk 19 above)
Comment (robinmolen) [reply src/components/designer/utils/fields.ts]: I think i've found what you meant. In the backend project, in the `FormStepDefinition.js`, we define a `componentNamespace`. Which is flat map of all form step components combined. This is similar to the `openForms.componentNamespace` i described above.

I think passing such a flat map the the FormBuilder (which passes it to these internal functions) should work fine. If we, at some point, require the context of the form steps, we can adjust this by passing the formSteps to the FormBuilder, and making the `componentNamespace` ourselves.

File: src/components/FormBuilder.tsx [hunk 30]
```diff
@@ -0,0 +1,72 @@
...
+export interface FormBuilderProps {
+  /**
+   * The components that are edited in the form builder.
+   */
+  components: AnyComponentSchema[];
+  /**
+   * A flatmap of all components in the form. Used when creating component keys to
+   * validate complete uniqueness.
+   */
+  componentNamespace: AnyComponentSchema[];
```
sergei-maertens, line 19: a flatmap would be `Record<str, AnyComponentSchema>`, here it's a collection of all components, so the docstring needs updating.

src/components/FormBuilder.tsx
sergei-maertens: I'm adding some comments in the code, because this implementation now is just wrong and tries to handle it in a generic way while it just can't do that:

* not every component type has a default value property
* default value types still *do* depend on the component definition itself
* this now assigns a default value for layout components too, and those components *do not hold values* (content, softrequirederrors, fieldset, columns)
* the renderer does implement this properly, otherwise it would break our SDK entirely

sergei-maertens: And w/r to overkill - the bundler should drop the dead code and only keep the used `extractInitialValues`, so I don't see this as an issue.

In the long term with the admin-ui, we want to show a styling-aware form preview anyway, and at that point, the renderer should just be used/embedded in the backend/formio-builder anyway because then you get a true representation with the proper NL DS components.

src/registry/index.ts
sergei-maertens, line 1: I'm not too fond of this random utilities being dropped in a module called `registry`, and a switch being necessary on the component type which spreads implementation details of each component type over multiple places again.

Why can't you extend the `RegistryEntry` type with an optional callback `getChildComponents` that returns `AnyComponentSchema[]` as signature, and then implement that callback for fieldset, editgrid and columns and wire them up? That's what we're doing with the edit form, default values, previews and the same pattern that we do in the renderer for extracting initial values, defining empty value, render implementation, visibility handling...

src/registry/index.ts
```diff
@@ -49,6 +49,25 @@ export const isKnownComponentType = (
...
+    case 'editgrid':
+    case 'fieldset': {
+      return component.components;
+    }
+    case 'columns': {
+      return component.columns.map(column => column.components);
+    }
+  }
+  return [];
+};
```
sergei-maertens: this looks like it's only needed to handle the columns, but why can't the columns just merge the component arrays from its columns rather than returning a nested array?

```tsx
return component.columns.map(column => column.components).flat(1);
```

src/components/FormBuilder.tsx
```diff
@@ -0,0 +1,72 @@
...
+  /**
+   * The regular component groups that are shown in the form builder components list.
+   */
+  groups: ComponentGroup[];
+  /**
+   * Custom-defined component presets that are shown in the form builder components list.
+   */
+  presets: PresetComponentDefinition[];
```
sergei-maertens: I kinda fail to see why these need to be props at this point, and why the storybook decorator with overridable params is necessary:

* you don't actually make use of these parameters anywhere, only use the defaults defined for tests
* at first instance, we'll just have our hardcoded list of groups and presets anyway, these can be defined as a constant in the formio-builder and made static

I think this removes a bunch of context and testing/stories complexity that's not used but does increase cognitive load and leads to confusion, because it creates the impression that it's used somewhere (why else would it be defined?).

The backend shouldn't have to define the groups/presets anyway, we can move that into the right place. Possibly in the future we'll allow preset definitions that are taken from the database, but then the prop is easily added. So let's cross that bridge when we get there.

sergei-maertens (changes requested):
I'm mostly approving this, but I do worry that the overall structure of the codebase is suffering a bit and we're in danger of making similar mistakes to what formio.js itself did, namely trying to solve everything from a base component perspective (they use a god class `Component`) rather than delegating to the registry for each component.

This is particularly dangerous when probing for component properties like `multiple`, `label`, `defaultValue` - things that *seem to* be similar concepts between component types, but aren't always. The most obvious example of that is the `components` property that holds child components, which has completely different semantics between `fieldset` and `editgrid` while the type is the exact same. Those dangers lurk around the corner for any component property, save for perhaps `id`, `key` and `label` (but even the latter already doesn't exist for the content component).

src/registry/index.ts
robinmolen: This was needed to place draggable items in the correct column in the previous setup. I assume this is still the case.

I can simplify it for now, and explore different options with the D&D layout implementation

src/registry/index.ts
robinmolen: That did not cross my mind, but sounds a lot cleaner..

I'm a bit hesitant with adding functionality to the `RegistryEntry`, trying to keep this as lean as possible. And the utility seemed like a okay solution.

I'm changing it around

src/components/FormBuilder.tsx
robinmolen: I though we discussed that adding the `presets` prop would be some preparation for that possible feature.
I think we also discussed something for the `groups`, but im not sure.

But, im fine with removing them now and simplifying the context/stories

src/components/FormBuilder.tsx
robinmolen: I don't fully agree with your statement, but i see how this wording could be confusing. I'll update the docstring

sergei-maertens (approved):
Alright, let's build on top of this :)

PR #295, opened 2026-06-09
Form designer drag-and-drop for layout components

### LINKED ISSUE #269: Form-designer - D&D inside layout components

robinmolen:
Closes #269

Extending the form designer drag-and-drop behavior by supporting multiple nested dropzones (i.e. layout components like editgrid, fieldset and columns).

Interaction testing seems rather flaky (related to #296, I think), so to prevent future headaches with having to re-run failing tests, I've opted to keep the storybook tests simple for now.

### DIFF
--- a/src/components/designer/Preview.tsx
+++ b/src/components/designer/Preview.tsx
@@ -11,11 +11,11 @@ import {hasOwnProperty} from '@/types';
 import './Preview.scss';
 import {DropZone, SortableComponent} from './dragDrop';
 import {getTargetDropzoneId} from './dragDrop/utils/dragTarget';
-import type {ComponentPlaceholder} from './types';
+import type {ComponentDefinition} from './types';
 import {COMPONENT_PLACEHOLDER_TYPE} from './types';
 
 export interface ComponentsPreviewProps {
-  components: (AnyComponentSchema | ComponentPlaceholder)[];
+  components: ComponentDefinition[];
   dropzoneId: string;
   hideEmptyMessage?: boolean;
 }
--- a/src/components/designer/Preview.tsx
+++ b/src/components/designer/Preview.tsx
@@ -1,3 +1,4 @@
+import {useDragOperation} from '@dnd-kit/react';
 import type {AnyComponentSchema} from '@open-formulieren/types';
 import clsx from 'clsx';
 import {FormattedMessage, useIntl} from 'react-intl';
@@ -9,6 +10,7 @@ import {hasOwnProperty} from '@/types';
 
 import './Preview.scss';
 import {DropZone, SortableComponent} from './dragDrop';
+import {getTargetDropzoneId} from './dragDrop/utils/dragTarget';
 import type {ComponentPlaceholder} from './types';
 import {COMPONENT_PLACEHOLDER_TYPE} from './types';
 
@@ -23,8 +25,13 @@ export const ComponentsPreview: React.FC<ComponentsPreviewProps> = ({
   dropzoneId,
   hideEmptyMessage = false,
 }) => {
-  const hasComponents =
-    components.filter(component => component.type !== COMPONENT_PLACEHOLDER_TYPE).length > 0;
+  const {target} = useDragOperation();
+  const targetDropzone = getTargetDropzoneId(target);
+
+  const isDraggingAboveDropzone = targetDropzone !== undefined && targetDropzone === dropzoneId;
+  // Count the number of components in the dropzone, excluding dragged components.
+  const componentCountThreshold = isDraggingAboveDropzone ? 1 : 0;
+  const hasComponents = components.length > componentCountThreshold;
 
   return (
     <ErrorBoundary>
--- a/src/components/designer/Preview.tsx
+++ b/src/components/designer/Preview.tsx
@@ -109,8 +109,7 @@ export const ComponentPreview: React.FC<ComponentPreviewProps> = ({component}) =
   const {
     preview: {designer: PreviewComponent},
   } = entry;
-  // @TODO Remove the undefined check when all components have a designer preview.
-  if (PreviewComponent === undefined || PreviewComponent === null) {
+  if (PreviewComponent === null) {
     console.info(`No designer preview found for component ${component.type}.`);
     return null;
   }
--- a/src/components/designer/dragDrop/utils/components.ts
+++ b/src/components/designer/dragDrop/utils/components.ts
@@ -1,23 +1,44 @@
 import type {AnyComponentSchema} from '@open-formulieren/types';
-import type {Draft} from 'immer';
-import {IntlShape} from 'react-intl';
+import type {IntlShape} from 'react-intl';
 
+import {
+  MAIN_DROPZONE_ID,
+  getComponentKeyFromDropzoneId,
+} from '@/components/designer/dragDrop/utils/dropzone';
 import type {ComponentPlaceholder} from '@/components/designer/types';
 import {COMPONENT_PLACEHOLDER_TYPE} from '@/components/designer/types';
 import {getRegistryEntry} from '@/registry';
 import {hasOwnProperty} from '@/types';
 
 type ComponentDefinition = AnyComponentSchema | ComponentPlaceholder;
-type DraftComponentDefinitions = Draft<{components: ComponentDefinition[]}>;
+
+const hasNestedChildren = (
+  children: AnyComponentSchema[] | AnyComponentSchema[][]
+): children is AnyComponentSchema[][] => children.length > 0 && Array.isArray(children[0]);
+
+interface IterComponentsResult {
+  /**
+   * The index of the current item.
+   */
+  index: number;
+  /**
+   * The current item.
+   */
+  component: ComponentDefinition;
+  /**
+   * The collection of items that the current item belongs to.
+   */
+  collection: ComponentDefinition[];
+}
 
 /**
  * Recursively (and depth-first) iterate over all components in the component definition.
  */
 function* iterComponents(
   componentDefinitions: ComponentDefinition[]
-): Generator<ComponentDefinition> {
-  for (const component of componentDefinitions) {
-    yield component;
+): Generator<IterComponentsResult> {
+  for (const [index, component] of componentDefinitions.entries()) {
+    yield {index, component, collection: componentDefinitions};
     if (component.type === COMPONENT_PLACEHOLDER_TYPE) continue;
 
     const {getChildComponents} = getRegistryEntry(component.type);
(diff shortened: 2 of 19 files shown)


src/components/designer/dragDrop/utils/components.ts
```diff
@@ -10,35 +14,110 @@ import {hasOwnProperty} from '@/types';
...
     if (!children.length) continue;
 
-    yield* iterComponents(children);
+    if (hasNestedChildren(children)) {
+      for (const child of children) {
+        yield* iterComponents(child);
+      }
+    } else {
+      yield* iterComponents(children);
+    }
```
sergei-maertens: I still don't like this construct for the same reasons as before - the array shape here is still tightly coupled to the `columns` component type, except now it's just very implicit. If in the future a different compont type gets added that doesn't have one or two levels, but three (purely hypothetical example: a table with columns inside each row could have `component.rows.columns.components`) which will require a `hasNestedNestedChildren` helper function.

Instead, you know that your generator is supposed to produce `IterComponentsReturn` - can't you make the registry callback for each component generate those? So instead of `getChildComponents`, implement `iterChildComponents*` and let that recurse. Each component type fully understands its own structure and can produce the necessary context.

src/components/designer/dragDrop/utils/components.ts
```diff
@@ -10,35 +14,110 @@ import {hasOwnProperty} from '@/types';
 type ComponentDefinition = AnyComponentSchema | ComponentPlaceholder;
 type DraftComponentDefinitions = Draft<{components: ComponentDefinition[]}>;
 
+const hasNestedChildren = (
+  children: AnyComponentSchema[] | AnyComponentSchema[][]
+): children is AnyComponentSchema[][] => children.length > 0 && Array.isArray(children[0]);
+
+interface IterComponentsReturn {
```
sergei-maertens: `Return` is a bad suffix for this name, because generators produce/yield results *and* can also return something and completes the generator (and can be of a different type even when calling `gen.return(...)`)

Maybe `IterComponentsResult` is a better name.

src/components/designer/dragDrop/utils/components.ts
```diff
@@ -10,35 +14,110 @@ import {hasOwnProperty} from '@/types';
...
+export const getDropzoneComponents = (
+  draft: DraftComponentDefinitions,
+  dropzoneId: string
+): ComponentDefinition[] | undefined => {
+  if (dropzoneId === MAIN_DROPZONE_ID) {
+    return draft.components;
+  }
+
+  const parentKey = getComponentKeyFromDropzoneId(dropzoneId);
+  return findDropzoneComponentsByKey(draft.components, parentKey);
```
sergei-maertens: I'm having a bit of trouble with the `key` aspect in this name, which creates the expectation that it points to a component key. That's not the case for columns, the index of the column is also included. Throughout the Open Forms' codebases, if you see a variable named `key`, 99% of the time it points to the exact key of a component and we should match that expectation.

Maybe `parentReference` is a better name?

src/components/designer/dragDrop/utils/components.ts
```diff
@@ -10,35 +14,110 @@ import {hasOwnProperty} from '@/types';
...
+    // The column index is part of the dropzone key, so we have to find the correct
+    // column.
+    if (hasNestedChildren(children)) {
+      for (const [index, child] of children.entries()) {
+        if (`${component.key}.${index}` === key) return child;
+      }
+    } else if (component.key === key) {
+      return children;
+    }
+  }
```
sergei-maertens: Here too I think the generator approach in the registry would clean up things, and you can then still collect the child components as soon as you have a match I think.

src/components/designer/dragDrop/utils/components.ts
```diff
@@ -107,74 +186,43 @@ export const createComponent = <S extends AnyComponentSchema>(
...
  */
 export const removeComponentFromDraft = (draft: DraftComponentDefinitions, key: string) => {
-  removeComponentFromComponents(draft.components, key);
+  for (const {index, component, collection} of iterComponents(draft.components)) {
+    if (component.type !== COMPONENT_PLACEHOLDER_TYPE && component.key === key) {
+      collection.splice(index, 1);
+      return;
+    }
+  }
 };
```
sergei-maertens: One thing I've also noticed while working on the msgspec branch is that it's quite annoying to work with a datastructure like the formio configuration/component definition that requires you to access the `.components` - here I'm seeing this too: the `draft` is passed, and you don't use anything else from the draft except diving directly into `draft.components`.

It greatly simplifies code in my experience to operate on a collection of components rather than on the container that holds this collection - the same function can be called in different contexts:

* pass the components of a formio configuration (`draft.components`)
* pass the components of a fieldset (`component.components`)
* pass the components of a columns column (`column.components`)
* do something on editgrids if necessary (`component.components`, but likely "multiplied" by the amount of items in the data)

This generally works better with static type checking as well, because you get the collection type check at the call site rather than needing to test the argument type (draft / configuration / fieldset or editgrid / columns which requires recursing into the column definitions)...

This also makes the `draft` and immer datastructure a top-level concern, allowing mutation in the utility functions, which also makes them maybe a bit more unit-test friendly. It kind of acts like a database transaction in your entrypoint.

src/components/designer/Preview.tsx
```diff
@@ -23,8 +25,12 @@ export const ComponentsPreview: React.FC<ComponentsPreviewProps> = ({
...
   hideEmptyMessage = false,
 }) => {
-  const hasComponents =
-    components.filter(component => component.type !== COMPONENT_PLACEHOLDER_TYPE).length > 0;
+  const {target} = useDragOperation();
+  const dropzone = getTargetDropzoneId(target);
+
+  const isDraggingAboveDropzone = dropzone === dropzoneId;
+  // Count the number of components in the dropzone, excluding dragged components.
+  const hasComponents = isDraggingAboveDropzone ? components.length > 1 : components.length > 0;
```
sergei-maertens:
```ts
// Count the number of components in the dropzone, excluding components currently being dragged.
const componentCountTreshold = isDraggingAboveDropzone ? 1 : 0;
const hasComponents = components.length > componentCountTreshold;
```

src/registry/types.ts
```diff
@@ -87,7 +87,10 @@ export interface RegistryEntry<S extends AnyComponentSchema> {
   // initial data
   defaultValue: DefaultValueOf<S> | JSONValue | undefined;
   comparisonValue?: React.FC<ComparisonValueProps>;
-  getChildComponents?: (component: S) => AnyComponentSchema[];
+  // All components that can have children should implement this method.
+  // The children should be returned as a one-level array, unless the nested structure is
+  // significant for the component (like in the case of a columns component).
+  getChildComponents?: (component: S) => AnyComponentSchema[] | AnyComponentSchema[][];
```
sergei-maertens: As remarked above, this works _coincidentally_ now because we only have components with either direct children or max 1 level of nesting (columns), but given the tree structure nature of components, you should be able to deal with arbitrary depth.

This *also* kind of makes the same mistake/assumption again that the meaning of `editgrid.components` is the same as `fieldset.components`, which it isn't. The data shape is the same, but the behaviour is completely different. For the form builder that's not relevant, the drag and drop and edit behaviour remains identical, but there's a risk here for other people learning the codebase that they get misled about this as well.

Deferring the implementation to each component type in the registry emphasizes that each component type should always be implemented and treated in its own context, as only there the meaning of a particular data structure is understood.

sergei-maertens (changes requested):

src/components/designer/dragDrop/utils/components.ts
robinmolen: The generator approach seems to work okay when the layout component has children. When there are no children, no instance from inside that layout component is returned. This causes problems when trying to place a component inside a empty layout component.

I have some idea's on how to handle this; ranging from separate iterate methods to returning half results for layout components without children.

robinmolen: The best solution i've found was to keep the generator outside the registry, and change te implementation of the `getChildComponents`.

## Short version:
A generator won't work for empty layout components. We can add a second generator that only returns the component collections, but this will result into a bunch of duplicate code and increased maintenance. Its a lot simpler to implement something similar to the current `getChildComponents`, with a consistent structure:
```
// Columns component
getComponentSlots: component =>
    component.columns.map((column, index) => ({
      reference: `${component.key}.${index}`,
      collection: column.components,
    }))
    
// Fieldset component
getComponentSlots: component => [
    {
      reference: component.key,
      collection: component.components,
    },
  ],
```

## Long story:
The problem is with layout components that don't have children.

With a `iterChildComponents` generator method we can iterate over the children of layout components. In some situations we need to know in which collection a component is placed (for when we want to move a specific component away from their current parent, when we replace the placeholder with an actual component). By extending the `yield` with collection and index data, we can provide this information. For when we search the children of a specific parent, we can reuse the `collection` and use `parentReference` as identifier:
```
iterChildComponents: function* (component, getRegistryEntry) {
    if (!component.columns) return;
    for (const [columnIndex, column] of component.columns.entries()) {
      for (const [childIndex, child] of column.components.entries()) {
        yield {
          index: childIndex,
          component: child,
          collection: column.components,
          parentReference: `${component.key}.${columnIndex}`,
        };

        // If any of its children has children as well, iterate over those as well
        const {iterChildComponents} = getRegistryEntry(child.type);
        if (iterChildComponents) {
          yield* iterChildComponents(child, getRegistryEntry);
        }
      }
    }
  }
  ```
The problem is, when a layout has no children, nothing will be yielded. This results in layout collections (which are used as dropzones for the DnD behavior) not being available.

Then we have two options:
* either return 'dummy' yields for empty layouts
* or implement a second (fairly similar) generator that returns a reference and the child components collection.

The first option will be a pain, having to deal with "is this yield a useable component, or just some dummy" in every implementation. So lets not do that.

The second option works, but will duplicate a lot of code and increase the amount of code to maintain:
```
iterComponentCollections: function* (component, getRegistryEntry) {
    if (!component.columns) return;

    for (const [index, column] of component.columns.entries()) {
      yield {
        component,
        collection: column.components,
        reference: `${component.key}.${index}`,
      };

      for (const child of column.components) {
        // If any of its children has children as well, iterate over those as well
        const {iterComponentCollections} = getRegistryEntry(child.type);
        if (iterComponentCollections) {
          yield* iterComponentCollections(child, getRegistryEntry);
        }
      }
    }
  }
  ```

So, i think a singular method is a lot simpler. This also limits the recursive code to only the `iterComponents` function, instead of spreading it across multiple registry entries.

+ bonus: it's very simple to implement for future layout components

sergei-maertens: Sounds good - the important part for me is that the component-specific processing code is in the registry and the generic code is actually generic and agnostic to the component type :)

src/components/designer/dragDrop/utils/components.ts
robinmolen: As i mentioned on the comment below, i've replaced the `getChildComponents` with a consistently shaped slots structure

sergei-maertens (approved):
Nice, this looks solid!

PR #300, opened 2026-06-25
[#6315] allow FAQ items to be configured through the form builder

### LINKED ISSUE #6315: Add support for configurable bottom-sheet tool tips
- [x] Implement bottom-sheet tool tip following the design in the front end  https://github.com/open-formulieren/formio-renderer/pull/334
- [x] Implement tool tip per field following the design in the front end https://github.com/open-formulieren/formio-renderer/pull/335
  - [x] Update field component schemas to include newly configurable tool tip related fields https://github.com/open-formulieren/types/pull/95
- [x] Allow multiple tool tips to be configured per field through the admin form builder https://github.com/open-formulieren/formio-builder/pull/300

**Additional notes for the form admin**
-  Allow bullet points & bold styles for WYSIWYG editor
- Configurable fields for the tool tips:
  - Title
  - Content

SonnyBA:
Allows form builders to configure FAQ items through the form builder.
Closes the last bullet point in https://github.com/open-formulieren/open-forms/issues/6315.

**TODO**

- [x] Pin types version to release that includes the `FAQItems` key for each component


src/components/formio/datefield.tsx
```diff
@@ -22,6 +23,7 @@ export const DateField: React.FC<JSX.IntrinsicElements['input'] & DateFieldProps
   label,
   required = false,
   tooltip = '',
+  faqElements = undefined,
```
viktorvanwijk: `undefined` is the default value when it's not passed
```suggestion
  faqElements,
```

src/components/formio/radio.tsx
```diff
@@ -110,6 +114,10 @@ export const Radio: React.FC<RadioProps> = ({
           />
         </button>
       )}
+
+      {faqItems.map(faqItem => (
+        <FAQItem faqItem={faqItem} />
+      ))}
```
viktorvanwijk: Why is this different from the other components, where the items are included using `{faqElements}`?

src/components/builder/rich-text.tsx
```diff
@@ -8,25 +8,27 @@
...
-  Required<(typeof ClassicEditor)['defaultConfig']>['fontColor']
+  Required<(typeof Editor)['defaultConfig']>['fontColor']
 >['colors'] extends Array<infer C>
   ? C
   : never;
 
 export interface RichTextProps {
   name: string;
   required?: boolean;
+  editor?: typeof ClassicEditor;
```
viktorvanwijk: Shouldn't this be the type accepted by the `CKEditor` component instead of the `ClassicEditor` we happen to inherit from?

viktorvanwijk (commented):
Few minor things :)

I have noticed that it's not possible to enter any text in the content field in storybook, though. Not sure if storybook is acting funky or if something needs to be adjusted. If I look at the `MultipleLanguages` story of the content component however, it does seem to work 🤔

src/components/CKEditor.tsx
```diff
@@ -30,7 +30,19 @@ import {
 import type {EditorConfig} from 'ckeditor5';
 import 'ckeditor5/ckeditor5.css';
 
-class Editor extends ClassicEditor {
+export class BasicEditor extends ClassicEditor {
+  public static override builtinPlugins = [Bold, List, Undo];
+
+  public static override defaultConfig: EditorConfig = {
+    toolbar: {
+      items: ['bold', '|', 'bulletedList', 'numberedList', '|', 'undo', 'redo'],
```
viktorvanwijk: This [comment](https://github.com/open-formulieren/open-forms/issues/5788#issuecomment-3840318137) lists paragraphs as well actually (probably `heading`?).

Image

src/components/builder/rich-text.tsx
SonnyBA: Tried using the CKEditor's base editor like so:
```
+++ w/src/components/builder/rich-text.tsx
@@ -6,18 +6,18 @@
  * Eventually replace this with something that is Slate.js based so that we can
  * incorporate our domain-specific variables/constructs for a better editing experience.
  */
-import {CKEditor} from '@ckeditor/ckeditor5-react';
+import {CKEditor as CKEditorComponent} from '@ckeditor/ckeditor5-react';
 import {AnyComponentSchema} from '@open-formulieren/types';
-import {ClassicEditor} from 'ckeditor5';
+import {Editor as CKEditor} from 'ckeditor5';
 import {useField, useFormikContext} from 'formik';
 import {useContext} from 'react';

-import Editor from '@/components/CKEditor';
 import {TemplatingHint} from '@/components/builder';
 import {Component, Description} from '@/components/formio';
 import {BuilderContext} from '@/context';

 import './richText.scss';
+import Editor from '../CKEditor';

 export type ColorOption = Required<
   Required<(typeof Editor)['defaultConfig']>['fontColor']
@@ -28,7 +28,7 @@ export type ColorOption = Required<
 export interface RichTextProps {
   name: string;
   required?: boolean;
-  editor?: typeof ClassicEditor;
+  editor?: typeof CKEditor;
   supportsBackendTemplating?: boolean;
   onChange?: (value: string) => void;
 }
@@ -54,7 +54,7 @@ const RichText: React.FC<RichTextProps> = ({
   const [props, , helpers] = useField<string>(name);
   return (
     <Component type={type} field={name} required={required} className="offb-rich-text">
-      <CKEditor
+      <CKEditorComponent
         editor={editor}
         config={{
           link: {
```

But Typescript wasn't happy about it:
```
Type 'typeof Editor | typeof Editor' is not assignable to type '{ create(...args: any): Promise<ClassicEditor>; EditorWatchdog: typeof EditorWatchdog;
ContextWatchdog: typeof ContextWatchdog; }'.
     Type 'typeof Editor' is not assignable to type '{ create(...args: any): Promise<ClassicEditor>; EditorWatchdog: typeof EditorWatchdog; ContextWatchdog: typeof
     ContextWatchdog; }'.
       The types returned by 'create(...)' are incompatible between these types.
         Type 'void' is not assignable to type 'Promise<ClassicEditor>'. 2322 | [typescript]
   ckeditor.d.ts:94:5: The expected type comes from property 'editor' which is declared here on type 'IntrinsicAttributes &
   IntrinsicClassAttributes<CKEditor<ClassicEditor>> & Readonly<Props<ClassicEditor>>'
```

src/components/CKEditor.tsx
SonnyBA: Updated the editor configuration to include basic textual operations only:
image

SonnyBA:
> Few minor things :)
> 
> I have noticed that it's not possible to enter any text in the content field in storybook, though. Not sure if storybook is acting funky or if something needs to be adjusted. If I look at the `MultipleLanguages` story of the content component however, it does seem to work 🤔

Not sure if this changed due to the recent changes but I could not reproduce this when I tested all the editable content field stories in storybook.

src/components/builder/faq-items.tsx
```diff
@@ -0,0 +1,124 @@
+import type {FAQItem} from '@open-formulieren/types/dist/common';
+import {FieldArray, useFormikContext} from 'formik';
+import {useContext} from 'react';
+import {FormattedMessage, useIntl} from 'react-intl';
+
+import {BuilderContext} from '@/context';
+
+import {MinimalEditor} from '../CKEditor';
+import {Panel, Tab, TabList, TabPanel, Tabs, TextField} from '../formio';
```
viktorvanwijk: I think it's a bit nicer to use the absolute import `@/components/...` here

src/components/builder/rich-text.tsx
viktorvanwijk: Hmm, OK. Perhaps they don't expose that specific type 🤔

src/components/builder/rich-text.tsx
```diff
@@ -37,7 +40,13 @@ export interface RichTextProps {
  * classic editor build, with some extra plugins enabled to match the features used/exposed
  * by Formio.js.
  */
-const RichText: React.FC<RichTextProps> = ({name, required, supportsBackendTemplating = false}) => {
+const RichText: React.FC<RichTextProps> = ({
+  name,
+  required,
+  editor = Editor,
+  supportsBackendTemplating = false,
+  onChange = undefined,
```
viktorvanwijk:
```suggestion
  onChange = () => {},
```
Then you can omit the question mark operator when calling it, because it's always a function :)

viktorvanwijk (commented):
> Not sure if this changed due to the recent changes but I could not reproduce this when I tested all the editable content field stories in storybook.

I see you've added `onChange` handlers to the components in `FAQItems`, so probably that has something to do with it :grin:. Can confirm that it works now!

Could you please respond to the comments I posted on two of the chromatic snapshots?

src/components/builder/faq-items.tsx
```diff
@@ -0,0 +1,124 @@
...
+                        name={`${arrayHelpers.name}[${index}].openForms.translations.${code}.label`}
+                        label={
+                          <FormattedMessage
+                            description="Label for translation message for FAQ label"
+                            defaultMessage="Label"
+                          />
+                        }
+                        onChange={event => {
+                          const {value: changedValue} = event.target;
+                          setFieldValue(`${arrayHelpers.name}[${index}].label`, changedValue);
```
SonnyBA: Reminder for myself, this currently sets the top-level attributes to the last changed value, while we probably only want to set the values from the "default" language.

viktorvanwijk (approved):
Seems OK to me, but probably good if Sergei takes a look as well :)

src/components/ComponentPreview.stories.tsx
```diff
@@ -30,6 +34,39 @@ export default {
 
 type Story = StoryObj<typeof ComponentPreview>;
 
+const faqConfiguration = {
```
sergei-maertens: please add a type here for future type checking safety, I'm fairly certain the inferred type will not be robust enough

```tsx
const faqConfiguration: {faqItems: FAQItemType[]} = {
```
should do

src/registry/addressNL/edit.tsx
```diff
@@ -29,6 +29,7 @@ import {
   Validate,
   useDeriveComponentKey,
 } from '@/components/builder';
+import FAQItems from '@/components/builder/faq-items';
```
sergei-maertens: please make sure `FAQItems` is exported from the `@/components/builder/index.ts` module, so you can move the import to the rest of the `'@/components/builder'` import group

src/registry/addressNL/edit.tsx
```diff
@@ -207,6 +209,7 @@ const EditForm: EditFormDefinition<AddressNLComponentSchema> = () => {
         <Key isManuallySetRef={isKeyManuallySetRef} generatedValue={generatedKey} />
         <Description />
         <Tooltip />
+        <FAQItems />
```
sergei-maertens: I'm wondering if we should move this into a separate tab entirely instead of adding this to the "Basic" tab. It's definitely not basic.

That would also allow you to get rid of the collapsible group.

src/components/CKEditor.tsx
```diff
@@ -30,7 +30,59 @@ import {
...
+    Table,
+    TableToolbar,
+    TextTransformation,
+    Undo,
+  ];
+
+  public static override defaultConfig: EditorConfig = {
+    toolbar: {
+      items: [
+        'heading',
```
sergei-maertens: We deliberately do *not* want to expose headings per the ticket specification - the markup nightmare for accessibility is something we want to avoid.

So, only `<p>` tags and inline formatting like bold/italic are allowed.

src/components/CKEditor.tsx
```diff
@@ -30,7 +30,59 @@ import {
...
+        'fontFamily',
```
sergei-maertens: not requested/specified


```suggestion
```

src/components/CKEditor.tsx
```diff
@@ -30,7 +30,59 @@ import {
...
+        'fontSize',
```
sergei-maertens:
```suggestion
```

not requested/specified

src/components/CKEditor.tsx
```diff
@@ -30,7 +30,59 @@ import {
...
+        'fontColor',
```
sergei-maertens:
```suggestion
```
not requested/specified

src/components/CKEditor.tsx
```diff
@@ -30,7 +30,59 @@ import {
...
+        'italic',
+        'bulletedList',
+        'numberedList',
+        '|',
+        'indent',
+        'outdent',
+        '|',
+        'alignment',
+        '|',
+        'blockQuote',
```
sergei-maertens:
```suggestion
```
not requested/specified

src/components/CKEditor.tsx
```diff
@@ -30,7 +30,59 @@ import {
...
+        'undo',
+        'redo',
+      ],
+    },
+    language: 'en',
+    licenseKey: 'GPL',
+    table: {
+      contentToolbar: ['tableColumn', 'tableRow', 'mergeTableCells'],
+    },
```
sergei-maertens: why is this here for the minimal editor?

src/components/CKEditor.tsx
```diff
@@ -30,7 +30,59 @@ import {
...
+    FontSize,
+    Heading,
+    Indent,
+    Italic,
+    List,
+    Paragraph,
+    Table,
+    TableToolbar,
+    TextTransformation,
+    Undo,
```
sergei-maertens: Surely we can remove many of these plugins that are not used in the minimal editor?

src/components/formio/faq-items.tsx
```diff
@@ -0,0 +1,42 @@
+import type {FAQItem as FAQItemType, SupportedLocales} from '@open-formulieren/types';
+import type React from 'react';
```
sergei-maertens: this import should not be necessary

src/components/formio/faq-items.tsx
```diff
@@ -0,0 +1,42 @@
...
+      if (!label) return accumulator;
+      accumulator.push(label);
+      return accumulator;
+    },
+    []
+  );
+  const selectedLabel = labels.length > 0 ? labels[0] : '';
+
+  return (
+    <div style={{paddingTop: '.375rem'}}>
```
sergei-maertens: can you please use a class name + CSS instead of inline styles? We have CSS build tooling :)

src/components/formio/faq-items.tsx
```diff
@@ -0,0 +1,42 @@
...
+      const translations = faqItem?.openForms?.translations?.[languageCode];
+      const label = translations?.label;
+
+      if (!label) return accumulator;
+      accumulator.push(label);
+      return accumulator;
+    },
+    []
+  );
+  const selectedLabel = labels.length > 0 ? labels[0] : '';
```
sergei-maertens: This... looks odd? Is that how we handle (for example) the `component.label` translations in the preview too?

In my mind the formio-builder always writes to the top-level field, and the "translations" tab writes to the language-specific fields.

I also see a translations UI embedded in the builder-side of things, whereas for other components all translations are managed on the "Translations" tab, see for example https://open-formulieren.github.io/formio-builder/?path=/story/public-api-componentconfiguration--radio

src/components/CKEditor.tsx
```diff
@@ -30,7 +30,59 @@ import {
...
+    },
+    language: 'en',
+    licenseKey: 'GPL',
+    table: {
+      contentToolbar: ['tableColumn', 'tableRow', 'mergeTableCells'],
+    },
+  };
+}
+
+export default class Editor extends ClassicEditor {
```
sergei-maertens: Please don't inline these exports, they're unreadable. Instead export at the bottom of the module:

```ts
export {MinimalEditor};
export default Editor;
```

src/components/builder/rich-text.tsx
```diff
@@ -8,27 +8,29 @@
...
-  >
+  Required<Required<(typeof Editor)['defaultConfig']>['fontColor']>['colors'] extends Array<infer C>
     ? C
     : never;
 
 export interface RichTextProps {
   name: string;
   required?: boolean;
+  label?: string;
+  editor?: typeof ClassicEditor;
```
sergei-maertens: In our own module we can define a union type:

```ts
export type AnyEditor = typeof MinimalEditor | typeof Editor;
```

and then you can 


```suggestion
  editor?: AnyEditor;
```
which seems to type check for me :)

src/components/builder/rich-text.tsx
```diff
@@ -38,44 +40,62 @@ export interface RichTextProps {
...
+  editor = Editor,
+  supportsBackendTemplating = false,
+  onChange = () => {},
+}) => {
   const {richTextColors} = useContext(BuilderContext);
   const {
     values: {type},
   } = useFormikContext<AnyComponentSchema>();
   const [props, , helpers] = useField<string>(name);
+  const htmlId = `editform-${name}`;
```
sergei-maertens: use `useId()` from React here instead, it guarantees unique IDs

src/components/builder/rich-text.tsx
```diff
@@ -38,44 +40,62 @@ export interface RichTextProps {
...
-                  rel: 'noopener noreferrer',
+    <Component
+      type={type}
+      field={name}
+      required={required}
+      label={label}
+      htmlId={htmlId}
+      className="offb-rich-text"
+    >
+      <div id={htmlId}>
```
sergei-maertens: this does nothing - clicking the label does not focus the editor, so there's no accessibility improvement.

Label/input IDs are used to related them together.

src/components/builder/rich-text.tsx
```diff
@@ -38,44 +40,62 @@ export interface RichTextProps {
...
-      {supportsBackendTemplating && <Description text={<TemplatingHint />} />}
+            fontColor: {
+              colors: richTextColors,
+            },
+          }}
+          data={props.value || ''}
+          onChange={(_, editor) => {
+            const html = editor.getData();
+            helpers.setValue(html);
+            onChange(html);
```
sergei-maertens: you probably added this to be able to manage translations etc., but those should be moved into the Translations tab. The built-in onchange + formik state update should really be sufficient.

sergei-maertens (changes requested):
I'm seeing quite a variation in the nature of changes in this PR - a single commit for the entire PR is not ideal. I'd expect separate commits for:

* rendering the faq items in component previews
* commit for the generic/general purpose faq item editing code
* commit that enables the faq item editing/configuration for the component types in the registry

sergei-maertens:
@SonnyBA I've rebased this branch on `main` (1.0.0)

src/registry/addressNL/preview.tsx
```diff
@@ -3,6 +3,7 @@ import {clsx} from 'clsx';
 import {FormattedMessage} from 'react-intl';
 
 import {Description, FieldSet, TextField} from '@/components/formio';
+import FAQItems from '@/components/formio/faq-items/faq-items';
```
sergei-maertens: here too you can do a nicer barrel export:


```suggestion
import {Description, FAQItems, FieldSet, TextField} from '@/components/formio';
```

src/components/builder/faq-items/i18n.tsx
sergei-maertens, line 1: I'm missing storybook stories (and docs) for this, to make it consistent with the other translations like here:

* https://open-formulieren.github.io/formio-builder/?path=/story/formio-builder-i18n-componenttranslations--default
* https://open-formulieren.github.io/formio-builder/?path=/story/formio-builder-values-translations--default

src/components/builder/faq-items/faq-items.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const fieldName = 'faqItems';
+  const {getFieldProps} = useFormikContext();
+  const {value: faqItems = []} = getFieldProps<FAQItem[] | undefined>(fieldName);
+
+  return (
+    <FieldArray name={fieldName}>
+      {arrayHelpers => (
+        <>
+          {faqItems.map((_, index) => (
+            <div className="faq-item" key={index}>
```
sergei-maertens: using the index as key will cause problems when the order of items is changed through the UI. we currently don't have this, but leaving this in will lead to unexpected bugs in the future.

I think we should probably add an `id` field to the type, and the builder can generate a UUID for it through `window.crypto`, so then we can use this ID as `key` value. 

Alternatively, we could compose a unique key through `${index}-${label}`, which would probably still work, until of course items with the same label are added :/ So I'd prefer the unique `id` property route.

src/components/builder/faq-items/faq-items.tsx
```diff
@@ -0,0 +1,76 @@
...
+              <TextField
+                name={`${arrayHelpers.name}[${index}].label`}
+                required={true}
+                label={
+                  <FormattedMessage
+                    description="Label for translation message for FAQ label"
+                    defaultMessage="Label"
+                  />
```
sergei-maertens, line 40:
```suggestion
                  <FormattedMessage
                    description="Label for FAQ item label configuration field"
                    defaultMessage="Label"
                  />
```

src/components/builder/faq-items/faq-items.tsx
```diff
@@ -0,0 +1,76 @@
...
+                }
+              />
+
+              <RichText
```
sergei-maertens: Can you add a `span` or something with label styling for the `"Content"` label?

src/components/builder/faq-items/faq-items.tsx
```diff
@@ -0,0 +1,76 @@
...
+                className="btn btn-secondary"
+                type="button"
+                onClick={() => arrayHelpers.remove(index)}
+              >
+                <i className="fa fa-times-circle-o" aria-hidden="true" />{' '}
+                <FormattedMessage
+                  description="'Remove item' screenreader button text"
+                  defaultMessage="Remove FAQ item"
+                />
+              </button>
```
sergei-maertens: The button has no vertical spacing with the "Add FAQ Item" button. Additionally, in the radio/select/selectboxes values it's a much smaller button with only an icon, which is inline with the item itself - I'd like a bit more consistent layout here.

We can do something like this:

```html
<div class="faq-item">
  <div class="form-group ... "> ... </div>  <!-- label field -->
  <div class="form-group ..."> ... </div> <!-- content editor -->
  <div class="faq-item__controls">
    <button> ... </button>
  </div>
</div>
```
with matching CSS:

```css
.faq-item {
  display: grid;
  grid-template-columns: 1fr max-content;
  grid-template-areas: 'label controls' 'content controls';

  .faq-item__controls {
    grid-area: controls;
  }
}
```

src/components/builder/faq-items/faq-items.tsx
```diff
@@ -0,0 +1,76 @@
...
+            </div>
+          ))}
+
+          <button
+            type="button"
+            className="btn btn-primary formio-button-add-another"
```
sergei-maertens: this `formio-button-add-another` class name doesn't have any styling anymore in our own css and is obsoleted

src/components/builder/faq-items/faq-items.tsx
sergei-maertens: Alternatively, a button styling approach like for the map overlays is also an option:

Image but that's rather an exception, as for the columns we also use the tabular structure:

Image

src/components/builder/faq-items/faq-items.tsx
```diff
@@ -0,0 +1,76 @@
...
+            onClick={() => {
+              arrayHelpers.push({
+                label: '',
+                content: '',
+              });
+            }}
+          >
+            <i className="fa fa-plus" aria-hidden="true" />{' '}
+            <FormattedMessage
+              description="'Add FAQ item' button text for 'multiple' FAQ items"
```
sergei-maertens:
```suggestion
              description="'Add FAQ item' button text"
```
not sure why the "multiple" is quoted here?

src/components/builder/faq-items/i18n.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+export function FAQItemsTranslations() {
+  const intl = useIntl();
+  const {activeLanguage} = useContext(ComponentTranslationsContext);
+  const {getFieldProps} = useFormikContext();
+  const name = 'faqItems';
+  const {value: faqItems = []} = getFieldProps<FAQItem[] | undefined>(name);
+
+  const {formType} = useContext(BuilderContext);
+  return formType === 'appointment' ? null : (
```
sergei-maertens: Bit weird that you can configure FAQ Items in appointment mode, but not the translations? See https://661e66a8eceef318207db8a5-pawmkjymvg.chromatic.com/?path=/story/public-api-componentconfiguration-appointment-mode--textfield

I don't see why FAQ Item translations would not be possible in appointment mode tbh.

src/components/builder/faq-items/i18n.tsx
```diff
@@ -0,0 +1,63 @@
...
+    <>
+      {faqItems.map((faqItem, index) => (
+        <div className="form-group" key={index}>
```
sergei-maertens: I don't think this layout/markup works - I'm thrown off by the lack of language tabs because this is suddenly no longer part of the existing tabular layout.

For example with the radio component:

Image

I would expect another table to be present here, with the heading "FAQ Item translations", then for each item two rows:

* Location: "FAQ Item n label", value: "<untranslated label value>", translations: translation field
* Location: "FAQ Item n content", value: "<untranslated content value>", translations: content editor

src/components/builder/i18n.stories.tsx
```diff
@@ -99,10 +136,26 @@ export const Default: Story = {
...
       const literal1Reference = canvas.getByText('Hi there {{ firstName }}');
       expect(literal1Reference).toBeVisible();
       await canvas.findByDisplayValue('Hallo daar, {{ firstName }}');
+
+      const faqItem1Labels = canvas.getAllByText('Should I fill this in?');
+      await expect(faqItem1Labels.length).toEqual(2);
+      await expect(faqItem1Labels[0]).toBeVisible();
+      const faqItem2Labels = canvas.getAllByText("I've XYZ, should I fill this in?");
+      await expect(faqItem2Labels.length).toEqual(2);
+      await expect(faqItem2Labels[0]).toBeVisible();
```
sergei-maertens, line 158: `await expect` doesn't do anything - `expect` is sync :)

PR #302, opened 2026-06-30
Add floating action buttons to form designer components

### LINKED ISSUE #270: Form-designer - Floating component actions
Implementing the floating component action buttons for changing the D&D position, editing the configuration (opening the modal) and removing the component from the form step.

robinmolen:
Closes #270 

This PR depends on #295

Introducing the component floating controls for editing the component configuration, and deleting the component from the form definition builder.

### DIFF
--- a/.storybook/decorators.tsx
+++ b/.storybook/decorators.tsx
@@ -3,7 +3,7 @@ import {Formik} from 'formik';
 
 import {ModalContext} from '@/components/Modal';
 import {ReferenceListsTableItem} from '@/components/builder/values/reference-lists/types';
-import {BuilderContext, DesignerContext, MapOverlayTileLayer, MapTileLayer} from '@/context';
+import {BuilderContext, MapOverlayTileLayer, MapTileLayer} from '@/context';
 import {
   DEFAULT_AUTH_PLUGINS,
   DEFAULT_COLORS,
@@ -142,18 +142,6 @@ export const BuilderContextDecorator: Decorator = (Story, context) => {
     >
       <Story />
     </BuilderContext.Provider>
-  );
-};
-
-export const DesignerContextDecorator: Decorator = (Story, context) => {
-  return (
-    <DesignerContext.Provider
-      value={{
-        componentNamespace: context.parameters?.designer?.componentNamespace || [],
-      }}
-    >
-      <Story />
-    </DesignerContext.Provider>
   );
 };
 
--- a/.storybook/decorators.tsx
+++ b/.storybook/decorators.tsx
@@ -172,3 +172,12 @@ export const withFormik: Decorator = (Story, context) => {
     </Formik>
   );
 };
+
+export const overrideWindowConfirm: Decorator = (Story, context) => {
+  window.confirm = () => {
+    context.parameters?.windowConfirm?.();
+    return true;
+  };
+
+  return <Story />;
+};
--- a/src/components/designer/FormioDefinitionDesigner.stories.tsx
+++ b/src/components/designer/FormioDefinitionDesigner.stories.tsx
@@ -10,7 +10,7 @@ import type {Meta, StoryObj} from '@storybook/react-vite';
 import {useState} from 'react';
 import {expect, fn, userEvent, waitFor, within} from 'storybook/test';
 
-import {overrideWindowConfirm, withFormik} from '@/sb-decorators';
+import {BuilderContextDecorator, overrideWindowConfirm, withFormik} from '@/sb-decorators';
 
 import FormioDefinitionDesigner from './FormioDefinitionDesigner';
 import type {FormioDefinitionDesignerProps} from './FormioDefinitionDesigner';
@@ -83,11 +83,17 @@ const StorybookFormioDefinitionDesigner = (props: FormioDefinitionDesignerProps)
 
 export default {
   title: 'Form designer/Form designer',
-  decorators: [withFormik],
+  decorators: [withFormik, BuilderContextDecorator],
   component: FormioDefinitionDesigner,
   render: StorybookFormioDefinitionDesigner,
   parameters: {
     modal: {noModal: true},
+    formik: {
+      wrapForm: false,
+    },
+    builder: {
+      enableContext: true,
+    },
   },
   args: {
     onChange: fn(),
@@ -248,6 +254,52 @@ export const ComponentInEditgrid: Story = {
   },
 };
 
+export const EditComponent: Story = {
+  args: {
+    components: [
+      {
+        id: 'textfield',
+        key: 'textfield',
+        type: 'textfield',
+        label: 'Textfield',
+      } satisfies TextFieldComponentSchema,
+    ],
+  },
+  play: async ({canvasElement, args}) => {
+    const canvas = within(canvasElement);
+
+    const textfieldComponent = canvas.getByTestId('sortable-item-textfield');
+    expect(textfieldComponent).toBeVisible();
+
+    // Start editing the textfield component
+    within(textfieldComponent).getByLabelText('Textfield').focus();
+    await userEvent.click(within(textfieldComponent).getByRole('button', {name: 'Edit component'}));
+
+    const modal = canvas.getByRole('dialog');
+    await waitFor(() => {
+      expect(modal).toBeVisible();
+    });
+
+    const labelField = within(modal).getByLabelText('Label');
+    await userEvent.clear(labelField);
+    await userEvent.type(labelField, 'Some cool textfield label');
+
+    await userEvent.click(within(modal).getByRole('button', {name: 'Save'}));
+
+    // The onChange should have been called with the updated components
+    expect(args.onChange).toHaveBeenCalledWith([
+      expect.objectContaining({
+        id: 'textfield',
+        key: 'textfield',
+        type: 'textfield',
+        label: 'Some cool textfield label',
+      }),
+    ]);
+
+    expect(canvas.getByText('Some cool textfield label')).toBeVisible();
+  },
+};
+
 export const DeleteComponent: Story = {
   args: {
     components: [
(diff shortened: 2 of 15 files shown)


src/components/designer/FormioDefinitionDesigner.tsx
```diff
@@ -182,11 +205,60 @@ const FormioDefinitionDesigner: React.FC<FormioDefinitionDesignerProps> = ({
...
+  onClose,
+}) => {
+  const {builderInfo} = getRegistryEntry(component.type);
+
+  // When the component is new, closing the modal without saving should remove it.
+  const closeModal = () => (isNew ? onRemove() : onClose());
+
+  return (
+    <Modal isOpen closeModal={closeModal}>
+      <ComponentEditForm
```
robinmolen, line 277: ~Needs to be replaced with ComponentConfiguration~
The FormBuilder component will fill the role of ComponentConfiguration. This will provider the BuilderContext, giving the component builder access to backend information as registration attributes, prefill plugins, tile layers, etc.

src/components/designer/FormioDefinitionDesigner.stories.tsx
```diff
@@ -235,3 +253,286 @@ export const ComponentInEditgrid: Story = {
...
+    const modal = canvas.queryByRole('dialog');
+    await waitFor(() => {
+      expect(modal).toBeVisible();
+    });
+
+    const labelField = within(modal as HTMLElement).getByLabelText('Label');
+    await userEvent.clear(labelField);
+    await userEvent.type(labelField, 'Some cool textfield label');
+
+    await userEvent.click(within(modal as HTMLElement).getByRole('button', {name: 'Save'}));
```
sergei-maertens:
```suggestion
    const modal = canvas.queryByRole('dialog')!;
    await waitFor(() => {
      expect(modal).toBeVisible();
    });

    const labelField = within(modal).getByLabelText('Label');
    await userEvent.clear(labelField);
    await userEvent.type(labelField, 'Some cool textfield label');

    await userEvent.click(within(modal).getByRole('button', {name: 'Save'}));
```
you only need to handle the potential `null`, so this non-null assertion is a better compromise than full blown downcasting.

`as` is forbidden, and there are exceptions yes, but then there must always be a comment explaining why there is no other way for future developers and reviewers.

sergei-maertens: Looking at this, I don't see why you wouldn't be able to do:

```ts
const modal = canvas.getByRole('dialog');
```
or
```ts
const modal = await canvas.findByRole('dialog');
```

the former is sync like `queryByRole`, but both only return/expect one item and the `getBy` variant doesn't have the potential `null` return value.

src/components/designer/FormioDefinitionDesigner.stories.tsx
```diff
@@ -149,9 +158,18 @@ export const AddComponentToDropzone: Story = {
...
     await dragTo(textareaDraggableItem, dropzone);
 
+    // The edit modal for the textarea component should be opened
+    const modal = canvas.queryByRole('dialog');
+    await waitFor(() => {
+      expect(modal).toBeVisible();
+    });
+
+    // Save the component
+    await userEvent.click(within(modal as HTMLElement).getByRole('button', {name: 'Save'}));
```
sergei-maertens: see below for the downcast remarks

src/components/designer/FormioDefinitionDesigner.stories.tsx
```diff
@@ -235,3 +253,286 @@ export const ComponentInEditgrid: Story = {
...
+    // Start editing the textfield component
+    within(textfieldComponent).getByLabelText('Textfield').focus();
+    await userEvent.click(within(textfieldComponent).getByRole('button', {name: 'Edit component'}));
+
+    const modal = canvas.queryByRole('dialog');
+    await waitFor(() => {
+      expect(modal).toBeVisible();
+    });
+
+    await userEvent.click(within(modal as HTMLElement).getByRole('button', {name: 'Remove'}));
```
sergei-maertens: same downcast remark here

src/components/designer/FormioDefinitionDesigner.stories.tsx
```diff
@@ -235,3 +253,286 @@ export const ComponentInEditgrid: Story = {
...
+    await userEvent.click(
+      within(editgridComponent).getAllByRole('button', {name: 'Edit component'})[0]
+    );
+
+    const modal = canvas.queryByRole('dialog');
+    await waitFor(() => {
+      expect(modal).toBeVisible();
+    });
+
+    await userEvent.click(within(modal as HTMLElement).getByRole('button', {name: 'Remove'}));
```
sergei-maertens: same downcast remark here

src/context.ts
```diff
@@ -109,15 +109,13 @@ BuilderContext.displayName = 'BuilderContext';
   Designer
  */
 export interface DesignerContextType {
-  /**
-   * A collection of all components in the form, created by flattening the form
-   * definitions.
-   */
-  componentNamespace: AnyComponentSchema[];
+  editComponent: (component: AnyComponentSchema) => void;
+  deleteComponent: (component: AnyComponentSchema) => void;
```
sergei-maertens, line 113: shouldn't these be named with an `on` prefix like all other event handlers, for consistency?

.storybook/decorators.tsx
```diff
@@ -184,3 +172,14 @@ export const withFormik: Decorator = (Story, context) => {
     </Formik>
   );
 };
+
+export const overrideWindowConfirm: Decorator = (Story, context) => {
+  window.confirm = () => {
+    if (context.parameters?.windowConfirm) {
+      context.parameters?.windowConfirm();
+    }
```
sergei-maertens:
```suggestion
    context.parameters?.windowConfirm.?();
```
does the same _I think_

src/components/designer/dragDrop/ComponentControls.tsx
```diff
@@ -0,0 +1,61 @@
...
+        onClick={event => {
+          event.preventDefault();
+          editComponent(component);
+        }}
+      >
+        <span className="sr-only">
+          <FormattedMessage
+            description="Form designer preview edit component button label"
+            defaultMessage="Edit component"
+          />
```
sergei-maertens: I think you can use the same description and message for the title and label, these should be identical anyway. Now you end up with two different translation entries that have to be kept in sync.


```suggestion
          <FormattedMessage
            description="Form designer preview edit component button title"
            defaultMessage="Edit component"
          />
```

src/components/designer/dragDrop/ComponentControls.tsx
```diff
@@ -0,0 +1,61 @@
...
+        onClick={event => {
+          event.preventDefault();
+          deleteComponent(component);
+        }}
+      >
+        <span className="sr-only">
+          <FormattedMessage
+            description="Form designer preview delete component button label"
+            defaultMessage="Delete component"
+          />
```
sergei-maertens:
```suggestion
          <FormattedMessage
            description="Form designer preview delete component button title"
            defaultMessage="Delete component"
          />
```

src/components/designer/FormioDefinitionDesigner.tsx
```diff
@@ -120,34 +129,133 @@ const FormioDefinitionDesigner: React.FC<FormioDefinitionDesignerProps> = ({
     const newComponent = createComponent(sourceData.componentType, componentNamespace, intl);
     setItems(draft => {
       replacePlaceholderWithComponent(draft.components, newComponent);
+      onChange(current(draft.components) as AnyComponentSchema[]);
```
sergei-maertens: why is the downcast necessary here? at least put a comment in the code please

src/components/designer/FormioDefinitionDesigner.tsx
```diff
@@ -120,34 +129,133 @@ const FormioDefinitionDesigner: React.FC<FormioDefinitionDesignerProps> = ({
...
+    });
+
+    // Open the modal for the new component.
+    openModal(newComponent, true);
+  };
+
+  const openModal = (component: AnyComponentSchema, isNew: boolean = false) => {
+    setComponentToEdit({component, isNew});
+  };
```
sergei-maertens: you now have a callback function that changes identity every render, and is passed down in the the React context value, meaning the context value changes every render, meaning that every consumer of the context will re-render

This would be good to wrap in `useCallback` to make sure we have stable references and avoid unnecessary renders.

src/components/designer/FormioDefinitionDesigner.tsx
```diff
@@ -120,34 +129,133 @@ const FormioDefinitionDesigner: React.FC<FormioDefinitionDesignerProps> = ({
...
+  const closeModal = () => {
+    setComponentToEdit(null);
+  };
 
-      // @TODO cleanup, kinda dirty
-      // At this point draft.components should only contain component definitions.
+  const updateComponent = (component: AnyComponentSchema, previousComponentKey: string) => {
+    setItems(draft => {
+      replaceComponent(draft.components, previousComponentKey, component);
+      onChange(current(draft.components) as AnyComponentSchema[]);
```
sergei-maertens: why the downcast?

and with this casting and pattern being repeated often, I think it would be a good idea to define a helper function that internalizes this and itself is type safe, while avoiding downcasts being littered all over the codebase

src/components/designer/FormioDefinitionDesigner.tsx
```diff
@@ -120,34 +129,133 @@ const FormioDefinitionDesigner: React.FC<FormioDefinitionDesignerProps> = ({
...
+    const isParent =
+      getComponentSlots && getComponentSlots(component).some(slot => slot.collection.length > 0);
+
+    if (
+      isParent &&
+      !confirm(
+        intl.formatMessage({
+          description: 'Form designer delete parent component confirmation message',
+          defaultMessage:
+            'Removing this component will also remove all of its children. Are you sure you want to do this?',
```
sergei-maertens:
```suggestion
            'Removing this component will also remove all of its children. Are you sure you want to continue?',
```
My feeling says this is the form we commonly use, but please double check (in particular in the backend repository), my mind might be betraying me.

src/components/designer/FormioDefinitionDesigner.tsx
```diff
@@ -120,34 +129,133 @@ const FormioDefinitionDesigner: React.FC<FormioDefinitionDesignerProps> = ({
...
+        })
+      )
+    ) {
+      return;
+    }
+
+    setItems(draft => {
+      removeComponent(draft.components, component.key);
       onChange(current(draft.components) as AnyComponentSchema[]);
```
sergei-maertens: see downcast remarks form before

src/components/designer/FormioDefinitionDesigner.tsx
```diff
@@ -120,34 +129,133 @@ const FormioDefinitionDesigner: React.FC<FormioDefinitionDesignerProps> = ({
...
-        <div className="col col-2">
-          <ComponentsList />
-        </div>
-        <div className="col col-10">
-          <ComponentsPreview components={items.components} dropzoneId={MAIN_DROPZONE_ID} />
+    <DesignerContext.Provider
+      value={{
+        editComponent: openModal,
+        deleteComponent,
+      }}
```
sergei-maertens: This object itself may even need to be `useMemo'd` to avoid excessive re-renders 🤔

src/components/designer/FormioDefinitionDesigner.tsx
```diff
@@ -120,34 +129,133 @@ const FormioDefinitionDesigner: React.FC<FormioDefinitionDesignerProps> = ({
...
         </div>
-      </div>
-    </DragDropProvider>
+        {componentToEdit && (
+          <ComponentEditModal
+            component={componentToEdit.component}
+            isNew={componentToEdit.isNew}
+            onSubmit={component => {
+              updateComponent(component, componentToEdit?.component.key);
+              closeModal();
```
sergei-maertens:
```suggestion
              updateComponent(component, componentToEdit.component.key);
              closeModal();
```
line 208 already establishes it's not `null` or `undefined`, right?

src/components/designer/FormioDefinitionDesigner.tsx
```diff
@@ -120,34 +129,133 @@ const FormioDefinitionDesigner: React.FC<FormioDefinitionDesignerProps> = ({
...
+            }}
+            onRemove={() => {
+              deleteComponent(componentToEdit?.component);
```
sergei-maertens:
```suggestion
              deleteComponent(componentToEdit.component);
```

src/components/designer/FormioDefinitionDesigner.tsx
sergei-maertens: yeah nice, this was the intent and separation in components when I started out with this repo :-)

sergei-maertens (changes requested):

src/components/designer/FormioDefinitionDesigner.stories.tsx
robinmolen: There where some flakiness issues before, but both `findByRole` and `getByRole` seem to work fine now..

src/context.ts
robinmolen: I considered that, but that naming felt weird. These functions start the component editing and deleting, instead of being reactions of the edit/delete functions.

Adding the `on` prefix makes them seem as reactions/callbacks, instead of the functions that start the proces.

.storybook/decorators.tsx
robinmolen: Yeah that works fine, but i think that the explicit if statement is more readable/easier to predict (definitely for dev's not too familiar with optional chaining)

src/components/designer/dragDrop/ComponentControls.tsx
robinmolen: Good point. Changing it 👍

src/components/designer/FormioDefinitionDesigner.tsx
robinmolen: I've added comments to all of these downcast places, and replaced the downcasting with a proper assertion

src/context.ts
sergei-maertens: Okay, I can see that - though I suppose you could argue that it's a callback to an event (clicking the edit icon, dropping the placeholder...). Okay to leave as is.

.storybook/decorators.tsx
sergei-maertens: That's an interesting take considering you use the optional chaining operator on `parameters` :stuck_out_tongue:

robinmolen: Hahah, okay got me there 😄

src/components/designer/FormioDefinitionDesigner.tsx
robinmolen: Looking at the backend translations, we use both 😃, but we use `continue` more often. `continue` also sounds a bit nicer, so i'll change it

src/components/designer/FormioDefinitionDesigner.tsx
robinmolen: Yeah thats not needed

src/components/designer/FormioDefinitionDesigner.tsx
```diff
@@ -121,33 +131,158 @@ const FormioDefinitionDesigner: React.FC<FormioDefinitionDesignerProps> = ({
...
+
+        // Placeholders are only used for the drag and drop functionality, so there should
+        // not be any in the collection at this point. Additionally, we should only send
+        // component schemas to the backend.
+        const components: ComponentDefinition[] = current(draft.components);
+        assertNoPlaceholders(components);
+        onChange(components);
+      });
+    },
+    [getRegistryEntry, setItems, removeComponent, assertNoPlaceholders, current]
```
sergei-maertens: `getRegistryEntry`, `assertNoPlaceholders` and `current` can be left out because they're static imports (only changing datastructures from props, state and context should be in dependency arrays)

It doesn't break to keep them in, but I suspect they'll get picked up by ESLint once #297 is merged

sergei-maertens (approved):

src/components/designer/FormioDefinitionDesigner.tsx
robinmolen: Ah yeah.. 👍

PR #304, opened 2026-07-14
Preparing FormBuilder for backend integration

### LINKED ISSUE #271: Form-designer - integrate into backend project
Integrate the new form designer into the backend project, and replace the FormIOBuilder implementation

robinmolen:
Closes #271

Depends on #302

Final features for the new FormBuilder:

* Implementing formio-renderer `extractInitialValues` method, instead of trying to determine the form fields initial values manually.
* Implementing BuilderContext on the FormBuilder component, providing all underlying component forms access to backend information
* Implementing `onComponentUpdated` and `onComponentDeleted` callbacks, needed to trigger backend changes (like updating variable keys in logic rules when changing the key of a component)

With these, the FormBuilder is ready to be used in the backend project, as replacement for the Formio FormBuilder.


src/components/designer/dragDrop/utils/components.ts
```diff
@@ -10,6 +11,28 @@ import {COMPONENT_PLACEHOLDER_TYPE} from '@/components/designer/types';
 import {getRegistryEntry} from '@/registry';
 import {hasOwnProperty} from '@/types';
 
+type onReplaceCallback = ({
```
sergei-maertens:
```suggestion
type OnReplaceCallback = ({
```

src/components/designer/dragDrop/utils/components.ts
```diff
@@ -10,6 +11,28 @@ import {COMPONENT_PLACEHOLDER_TYPE} from '@/components/designer/types';
...
+  replacedComponent,
+  newComponents,
+}: {
+  replacedComponent: AnyComponentSchema;
+  newComponents: AnyComponentSchema[];
+}) => void;
+
+type onPlaceholderReplaceCallback = ({
```
sergei-maertens:
```suggestion
type OnPlaceholderReplaceCallback = ({
```
for consistent casing in type names

src/components/designer/dragDrop/utils/components.ts
```diff
@@ -174,10 +197,23 @@ export const removePlaceholder = (components: ComponentDefinition[]) => {
...
-export const removeComponent = (components: ComponentDefinition[], componentKey: string) => {
+export const removeComponent = (
+  components: ComponentDefinition[],
+  componentKey: string,
+  onRemove?: OnRemoveCallback
+) => {
   for (const {index, component, collection} of iterComponents(components)) {
     if (component.type !== COMPONENT_PLACEHOLDER_TYPE && component.key === componentKey) {
-      collection.splice(index, 1);
+      const removed: AnyComponentSchema[] = collection.splice(index, 1) as AnyComponentSchema[];
```
sergei-maertens:
```suggestion
      const removed = collection.splice(index, 1);
      assertNoPlaceholders(removed);
```

src/components/designer/dragDrop/utils/components.ts
```diff
@@ -174,10 +197,23 @@ export const removePlaceholder = (components: ComponentDefinition[]) => {
...
+      if (removed.length > 0 && onRemove) {
+        const newComponents: ComponentDefinition[] = current(components);
```
sergei-maertens: This `current` is spreading through the codebase very rapidly and it concerns me - the immer documentation mentions its first class usage as a debug tool and says to use it sparingly.

Additionally, there's no indication whatsoever here that this function is working with something immer related and in previous PRs we made some effort to make it agnostic, so now there's tight coupling introduced again.

All of this is making me feel like the way useImmer is used is forced and not the best fit, because you are interested in more than just the next component definition (and that's what your `useImmer` tracks):

* the component (sub)tree definition
* when a component is updated, the original definition and the updated definition (and whether it's new or not)
* when a component is deleted, the component that was deleted

I think that some raw `useState` combined with raw `produce` will get you further, in pseudo code for the various cases:

**move placeholder**

```
newItems = produce(items, draft => {
  removePlaceholderFromDraftComponents()
  insertPlaceholderInNewLocation()
});
setState(newItems)
```

**move component**

```
newItems = produce(items, draft => {
  removeComponentFromDraft()
  insertComponentInNewLocation()
});
setState(newItems)
onChange.?(newItems)
```

**handleDragOver**

```
newItems = produce(items, draft => {
  removePlaceholderOfNewComponentWhenNotDraggingOverTarget()
  movePlaceholderOrComponent()
});
setState(newItems)
onChange.?(newItems) // unless it was only a placeholder
```

**handleDragEnd**

```
// for new components
newComponent = createNewComponent()
newItems = produce(items, draft => {
  replacePlaceholderWithComponent()
})
setState(newItems)
onChange.?(newItems, {type: 'add', component: component})
openModal(newComponent)
```

with this one I'm not sure that this is the best course of action - wouldn't it make sense to only dispatch the `onChange` to the parent once the component configuration is actually saved in the modal? Because cancelling a new component removes it again

**updateComponent**

```
previousComponent = lookupPreviousComponentKeyFromItems()
newItems = produce(items, draft => {
  replaceComponentInDraftWithPreviousComponentKey()
})
setState(newItems)
onChange.?(newItems, {type: 'update', previous: previousComponent, component: component})
```

**deleteComponent**

```
newItems = produce(items, draft => {
  locateAndRemoveDeleteComponent()
})
setState(newItems)
onChange.?(newItems, {type: 'remove', component: deletedComponent})
```

---

Note that I've also put all the event handling here in the `onChange` callback - that thing is going to fire every time anyway, so you feed it the updated component tree (always), and a second argument with the mutation event that holds event-specific data, which is a plain discriminated union.

This way, you don't need a single `current` call, and you get the references to the final immer state + none of your helper functions don't have to be aware of immer semantics at all

sergei-maertens (changes requested):

package.json
```diff
@@ -135,6 +138,8 @@
     "immer": "^11.1.8",
     "leaflet": "^1.9.4",
     "leaflet-draw": "^1.0.4",
+    "leaflet-geosearch": "^4.2.0",
+    "leaflet-gesture-handling": "^1.2.2",
```
sergei-maertens, line 142: I think these can go in `peerDependencies`?

package.json
```diff
@@ -67,13 +67,15 @@
...
     "@types/leaflet-draw": "^1.0.11",
+    "@types/leaflet-geosearch": "^2.7.1",
     "@types/lodash": "^4.17.0",
     "@types/proj4leaflet": "^1.0.10",
     "@types/react": "^18.2.45",
     "@types/react-modal": "^3.16.3",
     "@types/react-signature-canvas": "^1.0.5",
     "@types/sanitize-html": "^2.9.0",
     "@types/signature_pad": "^2.3.6",
+    "@utrecht/component-library-react": "14.0.1",
```
sergei-maertens: latest renderer depends on `14.0.2`


```suggestion
    "@utrecht/component-library-react": "^14.0.2",
```

src/utils/inputmask.ts
```diff
@@ -14,10 +14,52 @@
...
 
 export const applyInputMask = (textValue: string, mask: string) => {
   const options = {placeholderChar: '_'};
-  const result = conformToMask(textValue, Utils.getInputMask(mask), options);
+  const result = conformToMask(textValue, getInputMask(mask), options);
   return result.conformedValue;
 };
+
+// Copied from Formio.js 4.13.x - src/utils/utils.js
+const getInputMask = (
```
sergei-maertens: we can drop this entirely - our renderer doesn't support input masks 😉 they're an accessibility problem anyway

src/registry/number/edit-validation.spec.ts
```diff
@@ -1,29 +1,21 @@
-import {Components} from 'formiojs';
+import type {NumberComponentSchema} from '@open-formulieren/types';
 import {expect, test} from 'vitest';
 
+import {createComponent} from '@/components/designer/dragDrop/utils/components';
 import {dummyBuilderContext, dummyIntl} from '@/tests/test-utils';
 
 import schemaFactory from './edit-validation';
 
 test('A formio Number component passes validation', () => {
```
sergei-maertens, line 8: I think this test is now effectively obsolete, because it was specifically testing that whatever `Components.create(...)` produces works in our builder, as that's the source from the backend that gets fed into our dialog.

So, just delete the test.

src/registry/currency/edit-validation.spec.ts
```diff
@@ -1,29 +1,20 @@
-import {Components} from 'formiojs';
+import type {CurrencyComponentSchema} from '@open-formulieren/types';
 import {expect, test} from 'vitest';
 
+import {createComponent} from '@/components/designer/dragDrop/utils/components';
 import {dummyBuilderContext, dummyIntl} from '@/tests/test-utils';
 
 import schemaFactory from './edit-validation';
 
 test('A formio Currency component passes validation', () => {
```
sergei-maertens, line 8: same here, just delete the test

sergei-maertens (changes requested):

src/utils/inputmask.ts
robinmolen: Mhh yeah.. i was thinking in case users/form builders define them. But yeah if the renderer doesn't support it, it doesn't really matter 😅

src/components/designer/dragDrop/utils/components.ts
robinmolen: Ah dumb... fixing it 🙈

src/components/designer/dragDrop/utils/components.ts
robinmolen: Ah shit

src/utils/inputmask.ts
sergei-maertens: I don't even think `@open-formulieren/types` defines them anymore even

package.json
robinmolen: This breaks storybook (in local dev, but i assume it would also break for CI builds)

robinmolen: Like discussed in the office, we can fix this breakage when formiojs has been removed as dependency. I'll leave this as is for now, and will move them in the "remove formiojs" branch

src/components/designer/dragDrop/utils/components.ts
robinmolen: Okay, i've made some changes to make this possible. In broad lines:

1. We only persist components that actually have been added. (So when adding a new component, we only fire the `onChange` on edit modal save, not before).
2. When adding a new component, we replace the placeholder AFTER the edit modal has been closed. This means deleting/canceling the edit modal will simply remove the placeholder, and saving the modal will replace the placeholder with the newly added component.
3. I've changed the internal `onChange` to the structure you suggested, and updated to the interaction tests to validate this. I've kept the external `onChange` (the `onChange` on the FormBuilder component) as it was before. I think it makes sense to keep the communication to the outside world as explicit as we can. But i'm open for change, if you think outside communication should follow the `onChange.?(newItems, {type: [action], component: component})` structure.
4. I've replaced useImmer with useState + produce, and removed immer logic away from the helper functions.
5. I've moved some more indepth logic to the helper functions, to keep the `FormioDefinitionDesigner` a bit dumber

src/components/designer/FormioDefinitionDesigner.tsx
```diff
@@ -37,10 +37,29 @@ const getData = (
   prop: Draggable<SortableItemData | DraggableMenuItemData | Data> | Droppable | null
 ): SortableItemData | DraggableMenuItemData | Data => prop?.data ?? {};
 
+type ComponentCreated = {
+  type: 'create';
+  component: AnyComponentSchema;
+  originalComponent: AnyComponentSchema;
```
sergei-maertens: discussed in the office - this is only present for some legacy backend callbacks/event handlers that we're not pleased with anyway, so this property can be dropped :)

src/components/designer/FormioDefinitionDesigner.tsx
```diff
@@ -37,10 +37,29 @@ const getData = (
...
+  component: AnyComponentSchema;
+};
+
+type ComponentEvent = ComponentCreated | ComponentUpdated | ComponentDeleted;
+
 export interface FormioDefinitionDesignerProps {
   components: AnyComponentSchema[];
   componentNamespace: AnyComponentSchema[];
-  onChange: (components: AnyComponentSchema[]) => void;
+  onChange: (components: AnyComponentSchema[], event?: ComponentEvent) => void;
```
sergei-maertens, line 55:
```suggestion
  onChange: (components: AnyComponentSchema[], event: ComponentEvent) => void;
```
I think a callback that only accepts one argument still matches this type definition, and it helps ensure call-sites that they're always going to get the `event` from us

src/components/designer/FormioDefinitionDesigner.tsx
```diff
@@ -37,10 +37,29 @@ const getData = (
...
 }
 
 const FormioDefinitionDesigner: React.FC<FormioDefinitionDesignerProps> = ({
```
sergei-maertens, line 58: Can you also add some doc blocks and specifically point out that the `components` prop is only used as initial state value and later updates will be ignored?

Maybe even rename it to `initialComponents`

src/components/designer/FormioDefinitionDesigner.tsx
```diff
@@ -95,16 +106,17 @@ const FormioDefinitionDesigner: React.FC<FormioDefinitionDesignerProps> = ({
 
     // When moving the placeholder outside the dropzone, remove it.
     if (!target && isNewComponent) {
-      setItems(draft => {
-        removePlaceholder(draft.components);
+      const newItems = produce(items, draftItems => {
+        removePlaceholder(draftItems);
       });
```
sergei-maertens, line 104:
```suggestion
      const newItems = produce(items, removePlaceholder);
```
is possible, but I'm not actually convinced this is more readable

sergei-maertens (changes requested):

src/components/designer/FormioDefinitionDesigner.tsx
robinmolen: `onChange` is also used to signal about position changes (moving one component below another), which doesn't require any additional handling besides saving the new structure. So i think we should keep the event optional? Otherwise, we would have to introduce a "move" event, which just kinda.. exists to be ignored 😄

src/components/designer/FormioDefinitionDesigner.tsx
robinmolen: Mhh, yeah.. i think im gonna leave it as is 🤔

robinmolen:
I noticed one addition change that makes the implementation to the backend easier. We already define a `uniquifyKey` in the backend, which is passed to the formio-builder and used in the component edit forms. We can use this as well when creating the components as part of the drag-n-drop behaviour.

This results in:
- Removing the `componentNamespace` prop from `FormBuilder` and `FormioDefintionDesigner` components
- Removing some duplicate code from the `components.ts` helper functions

src/components/designer/FormioDefinitionDesigner.tsx
robinmolen: the `originalComponent` has been removed from the `ComponentCreated` event

src/components/designer/FormioDefinitionDesigner.tsx
robinmolen: Like we discussed in the office, the `FormBuilder` component now also passes the component events along using the `onChange` method.
I've kept `event` optional for now

src/registry/number/edit-validation.spec.ts
robinmolen: Deleting this to a separate "drop formiojs" PR

src/registry/currency/edit-validation.spec.ts
robinmolen: Deleting this to a separate "drop formiojs" PR

src/components/designer/FormioDefinitionDesigner.tsx
sergei-maertens: ah yeah, the move makes sense indeed!

sergei-maertens (approved):

PR #306, opened 2026-07-21
Drop formio styles dependency

### LINKED ISSUE #305: Drop dependency on formio CSS (modal styles)
To rip out the formiojs dependency, we need a working solution for the modal styles/markup to prevent that from breaking down.

sergei-maertens:
Closes #305

**TODO**

- [x] Test this setup in a local build to see if it still integrates well with the backend

### DIFF
--- a/.storybook/decorators.tsx
+++ b/.storybook/decorators.tsx
@@ -38,20 +38,15 @@ export const ModalDecorator: Decorator = (Story, context) => {
         <Story />
       </ModalContext.Provider>
     );
+  // See src/components/Modal.tsx for the markup/classnames
   return (
-    <div
-      className="formio-dialog formio-dialog-theme-default component-settings"
-      style={{position: 'relative'}}
-    >
-      <div className="formio-dialog-overlay" style={{position: 'relative'}}></div>
-      <div className="formio-dialog-content">
-        <div></div>
-        <button
-          aria-label="close"
-          className="formio-dialog-close float-right btn btn-secondary btn-sm"
-        ></button>
-        <div className="component-edit-container">
-          <Story />
+    <div className="offb-modal">
+      <div className="ReactModal__Overlay ReactModal__Overlay--after-open offb-modal__overlay">
+        <div className="ReactModal__Content ReactModal__Content--after-open offb-modal__content">
+          <button aria-label="close" className="offb-modal__close btn btn-secondary btn-sm" />
+          <div className="component-edit-container">
+            <Story />
+          </div>
         </div>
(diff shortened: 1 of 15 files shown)


sergei-maertens:
Verified that this update does not appear to cause regressions in our current backend, which still renders the "normal" formio modal now, but the crux is that the backend itself loads the necessary styles.

Once formio.js is completely ripped out of the formio-builder, the backend can use the new form definition editor component which uses our own dialog/modal, and that will be the real test I suppose :) So I don't rule out further tweaks after merging this, but it looks to me that this *can* be merged safely.

.storybook/decorators.tsx
```diff
@@ -38,20 +38,17 @@ export const ModalDecorator: Decorator = (Story, context) => {
...
-          aria-label="close"
-          className="formio-dialog-close float-right btn btn-secondary btn-sm"
-        ></button>
-        <div className="component-edit-container">
-          <Story />
+    <div className="offb-modal">
+      <div className="ReactModal__Overlay ReactModal__Overlay--after-open offb-modal__overlay">
+        <div className="ReactModal__Content ReactModal__Content--after-open offb-modal__content">
+          {/* Formio drops in an empty div for whatever reason :) */}
+          <div />
```
robinmolen: I don't think this empty div has any meaning (removing it doesn't seem to have impact on any styling/function). Its also no-longer part of the actual Modal component, so outside this empty div only exists in/for storybook.

I think it was added by `overlayElement`, which has now been removed:
```
overlayElement={(props, contentElement) => (
        <>
          <div {...props}></div>
          {contentElement}
        </>
      )}
```
So i think we can safely remove it?

robinmolen (changes requested):
Codewise, looks great! There are a few changes highlighted by chromatic that need fixing :)

.storybook/decorators.tsx
sergei-maertens: This still happens in the backend integration, but yeah I expect that deleting it has no adverse side-effects. Will take a look at the Chromatic remarks!

src/formio.scss
```diff
@@ -0,0 +1,38 @@
...
+}
+
+.btn-xxs {
+  padding: 2px;
+  font-size: 10px;
+  line-height: 1.2em;
+  border-radius: 0;
+  width: 18px;
+  height: 18px;
+}
```
robinmolen, line 38: Dank ❤️

robinmolen (approved):
Happy happy, lets drop formio

PR #308, opened 2026-07-22
🔥  Drop formiojs dependencies

### LINKED ISSUE #307: Drop dependency Formiojs
Remove the Formiojs dependency from the projects, and replace whatever we still use with local implementations

- [x] Formio-builder #308 
- [x] Backend project

When removing formiojs from the backend, also add a end-to-end to validate that open-formulieren/open-forms#6297 doesn't happen anymore

robinmolen:
Closes #307
Depends on #306 

Replacing the remaining formiojs functionalities in the formio-builder with custom implementations, and dropping the formiojs dependency.

As inputMasks are no-longer supported by the formio-renderer, the formio-builder will also not provide this functionality anymore

~Marked as draft as #306 needs to be merged before this one~


package.json
robinmolen, line 141: Discussion from https://github.com/open-formulieren/formio-builder/pull/304#discussion_r3620905029

I tried moving this to the peerDependencies, but still got hit with "could not find imports bla bla bla". NPM had, at some point, an `auto-install-peers` but this no-longer exists?

Leaving this here for now

src/components/designer/dragDrop/utils/components.ts
```diff
@@ -11,36 +11,57 @@ import {COMPONENT_PLACEHOLDER_TYPE} from '@/components/designer/types';
...
+    const {getComponentSlots, holdsData} = getRegistryEntry(component.type);
+    yield {index, component, path, collection: componentDefinitions, holdsData};
 
-    const {getComponentSlots} = getRegistryEntry(component.type);
     if (!getComponentSlots) continue;
 
     for (const slot of getComponentSlots(component)) {
-      yield* iterComponents(slot.collection);
+      yield* iterComponents<S>(
+        slot.collection as S[],
```
robinmolen: Im gonna take a look if i can remove this `as`

robinmolen: This ties together with `ComponentDefinition`, which is used for the FormioDefinitionDesigner. This makes it so that components could have a temporary component definition that is not defined in `AnyComponentSchema`. In reality this temporary component is only added at a brief time when interacting with the FormioDefinitionDesigner, but this has effect on all definitions of children components.

src/components/builder/key.tsx
```diff
@@ -1,17 +1,27 @@
...
 import {TextField} from '../formio';
 
-const useGenerateUniqueKey = (component: ExtendedComponentSchema): string => {
+const getKeySeed = (component: AnyComponentSchema): string => {
+  for (const key of ['title', 'label', 'placeholder']) {
+    if (hasOwnProperty(component, key) && !!component[key]) {
+      return component[key] as string;
+    }
+  }
+  return component.type;
```
sergei-maertens: we don't have `title` anywhere:

```
➜  open-forms-types git:(main) rg title
➜  open-forms-types git:(main) 
```

which reduces the loop to two items, and at this point we can use an explicit if/else branch with type safety without needing a downcast - it required some tweaking which I've done in 56d292da911855097759d4e5b9bcc12028617814 😄

src/components/builder/simple-conditional.tsx
```diff
@@ -15,6 +14,20 @@ import {hasOwnProperty} from '@/types';
 import {Panel, Select} from '../formio';
 import ComponentSelect from './component-select';
 
+// Copied from formiojs/types/components/schema.d.ts
+interface ConditionalOptions {
```
sergei-maertens: Why not use the one that's actually used in our types? https://github.com/open-formulieren/types/blob/ab1adba60b3dc7fdd8599f50a853e87f4c0508ee/src/extensions.ts#L57

The formiojs conditional is notably broken - the `eq` always operates from a string type, but we use the proper primitives.

Additionally, we don't support `json` at all, so you should just import our own `Conditional` type :-)

src/components/builder/simple-conditional.tsx
```diff
@@ -1,11 +1,10 @@
 import type {AnyComponentSchema} from '@open-formulieren/types';
 import {useFormikContext} from 'formik';
-import {Utils as FormioUtils} from 'formiojs';
-import type {ConditionalOptions} from 'formiojs';
 import {useContext, useEffect} from 'react';
 import {FormattedMessage} from 'react-intl';
 import {usePrevious} from 'react-use';
 
+import {findComponent} from '@/components/designer/dragDrop/utils/components';
```
sergei-maertens: this has nothing to do with drag and drop anymore, so I suggest moving the definition of these general-purpose utilities in a better location.

In the renderer we have a top-level `src/formio.ts` module that would be nice to mimick (structure-wise) here I think.

src/components/designer/dragDrop/utils/components.ts
```diff
@@ -29,18 +33,28 @@ interface IterComponentsResult {
 /**
  * Recursively (and depth-first) iterate over all components in the component definition.
  */
-function* iterComponents(
-  componentDefinitions: ComponentDefinition[]
+export function* iterComponents(
+  componentDefinitions: ComponentDefinition[],
+  parentPath: string = ''
```
sergei-maertens: I find `path` a very confusing name, because `components.0.components.2.columns.1.components.5` is also a path :)

In the backend terms like [`parent_key`](https://github.com/open-formulieren/open-forms/pull/6444/changes#diff-d35129e6e728778dbc07c210e5d41c7c013cee4a1568afebf7bb96b279eea5a1R44) are used, or `data_path`.

I think here a name like `parentKeysPrefix` would better cover the meaning while being explicit about the source of the values/segments.

src/registry/types.ts
```diff
@@ -2,6 +2,7 @@ import type {AnyComponentSchema, JSONValue} from '@open-formulieren/types';
 import type {IntlShape, MessageDescriptor} from 'react-intl';
 import type {z} from 'zod';
 
+import type {COMPONENT_PLACEHOLDER_TYPE} from '@/components/designer/types';
```
sergei-maertens: isn't it weird that the type is defined in this module yet you import the constant (that arguable decides the type) from a more narrow place?

The docstring also describes that it's a placeholder used in the Form Designer context, so perhaps the `ComponentPlaceholder` really belongs where it was originally.

I don't really see how `ComponentDefinition` or `ComponentPlaceholder` relate to the registry concept of our particular component types - they're generic types that sit on top/above the registry and rely on the `AnyComponentSchema` registry abstraction.

Maybe it's `ComponentSlot` that needs to move into the `designer` types rather than the other way around?

I'm even thinking that a top-level `src/designer` folder or module could work here, for the non-component code like types, but `src/components/designer/types` probably is good enough too.

src/components/builder/component-select.tsx
```diff
@@ -19,22 +21,25 @@ export type ComponentSelectProps = SelectProps<ComponentOption, false> & {
...
-  const {values} = useFormikContext<ExtendedComponentSchema>();
+  const {values} = useFormikContext<AnyComponentSchema>();
 
   // Get all the components in the form from Formio
   const options = props.options || [];
   if (!props.options) {
-    FormioUtils.eachComponent(
-      getFormComponents(),
-      (component: ExtendedComponentSchema, path: string) => {
-        // FIXME: calculate path of the component properly instead of just using the key
```
sergei-maertens, line 30: I think this FIXME needs to be kept?

The detection here is to prevent the component from being listed itself, but in editgrids this is different because the component key is different from it's local editgrid path which is prefixed with the editgrid component key.

This *would* cause problems if you have a component key that's duplicated outside and inside the edit grid, which is okay for Formio - our backend now still enforces that this is not allowed, but some rework done recently is bringing us ever-closer to being able to drop that requirement. For that reason, I want to keep the comment.

sergei-maertens (changes requested):

src/components/builder/simple-conditional.tsx
robinmolen: Mhh, yeah alright. I thought it would be best to keep this close to the other components helper methods.

Im tempted to move `iterComponents` also to `src/formio.ts`, as its now no-longer just a drag-n-drop feature.

src/registry/types.ts
robinmolen: Like discussed via slack, i've reverted the `ComponentSlot` type back to strict. All slots now provide `ComponentDefinition[]` collections, as its possible to encounter a `ComponentPlaceholder` in the slot collections. The `isPlaceholder` guard can be used to handle situations where only `AnyComponentSchema` components need to be handled.

src/components/builder/simple-conditional.tsx
sergei-maertens: You have my blessing ;)

src/components/builder/component-select.tsx
robinmolen: Ahh damn, i thought this wasn't an issue anymore. The previous condition compared the path with the component key, which will only work if the component is not in an editgrid (you can see this bug in action in the test environment. Add a textfield to an editgrid, open de simple logic tab, and you can select the current component).

If we would want a full path comparison, I think we could create a full data path as we render the components tree, and pass that information along when opening the edit modal.

sergei-maertens: yeah, but out of scope for now :) I'd really like to wrap up the backend data structures first and then see if we can use a similar setup in the frontend, instead of arriving at two parallel implementations again

src/components/builder/key.tsx
sergei-maertens: I think you force pushed over my commit that fixed this 😬

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #309, opened 2026-07-29
Always grab the builder info from the registry

### LINKED ISSUE #292: Update `ComponentConfiguration` to grab the `builderInfo` from the component registry
Currently this must be passed as an explicit prop, but the recent components list development stores the builder info directly in the registry, allowing us to avoid having to pass that prop explicitly to the `ComponentConfiguration` component.

Instead, the `ComponentEditForm` (underlying component) should pull that from the registry when it receives the `component` prop. That can and will clean up a lot of repetition in stories.

sergei-maertens:
Closes #292

This should never be passed as prop anymore because we have all the metadata for our known component types in our component registry, which is the single source of truth.

Note that we can also remove the documentation, weight and group properties from the type since those are unused anyway.


robinmolen (approved):
Nice cleanup 🔥 Lets merge it!

PR #310, opened 2026-07-29
Fix up dependencies

sergei-maertens:
Formiojs required us to use legacy-peer-deps, which (turns out) created quite the mess :thisisfine:


vite.config.ts
```diff
@@ -55,10 +56,13 @@ export default defineConfig(({mode}) => ({
...
     preprocessorOptions: {
       scss: {
+        api: 'modern',
         quietDeps: true,
         // silence bootstrap @import statements
         silenceDeprecations: ['import'],
-        importers: [new sass.NodePackageImporter()],
+        // due to both sass and sass-embedded being installed, these types get messed
+        // up
+        importers: [new NodePackageImporter()] as unknown as _NodePackageImporter[],
```
sergei-maertens, line 65: This is cursed

package.json
robinmolen: Can we also move `@open-formulieren/formio-renderer` to the peer dependencies?

robinmolen (approved):
One question, but besides that it looks good

package.json
sergei-maertens: I will try it once CI is green with the blur/focus fixes!

sergei-maertens: Seems to be fine, let's see what CI thinks

